PR #1, opened 2016-11-23
Cleanup & a few bug fixes regarding method re-implementation

iamahuman:
- Removed resolution to wrong symbol `artQuickGenericJniTrampoline` (not a quick code, just inner impl.)
  - Allow ArtMethod instance mutation via kFastNative flag (needed by method restoration when calling the original method)
  - Sepearation between allocation ($alloc, corresponds to new-instance or Unsafe.allocateInstance) and initialization ($init, coressponds to &lt;init&gt;) - allows overriding of constructors
  - Selective patching and restoration of individual ArtMethod fields
  - Corrected a bug causing stack frame recognition of re-implemented methods to fail due to incorrect resolution of quick_generic_jni_trampoline

Suggestions are welcome!

### DIFF
--- a/index.js
+++ b/index.js
@@ -53,6 +53,7 @@ const kAccPublic = 0x0001;
 const kAccStatic = 0x0008;
 const kAccFinal = 0x0010;
 const kAccNative = 0x0100;
+const kAccFastNative = 0x00080000;
 
 // jobject reference types
 const JNIInvalidRefType = 0;
@@ -121,9 +122,9 @@ function Runtime() {
     this.synchronized = function (obj, fn) {
         assertCalledInJavaPerformCallback();
 
-        const objHandle = obj.hasOwnProperty('$handle')? obj.$handle : obj;
+        const objHandle = obj.hasOwnProperty('$handle') ? obj.$handle : obj;
         if (!(objHandle instanceof NativePointer)) {
-            throw new Error("It's neither a pointer nor a Java instance");
+            throw new Error("Java.synchronized: the first argument `obj` must be either a pointer or a Java instance");
         }
 
         const env = vm.getEnv();
@@ -131,10 +132,12 @@ function Runtime() {
         try {
             fn();
         } finally {
-            try {
-                checkJniResult("VM::MonitorExit", env.monitorExit(objHandle));
-            } catch (e) {
+            const result = env.monitorExit(objHandle);
+            if (result != JNI_OK) {
                 env.checkForExceptionAndThrowIt();
+
+                // If there were no exception, throw the JNI's one instead.
+                throw new Error("VM::MonitorExit failed: " + result);
             }
         }
     };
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -2995,66 +3135,241 @@ function Env(handle) {
...
+    this._current = specByPointerSize && specByPointerSize[Process.pointerSize];
+    return this._current;
+}
 
-        let jniCodeOffset = -1;
-        let accessFlagsOffset = -1;
-        let remaining = 2;
-        for (let offset = 0; offset !== 64 && remaining !== 0; offset += 4) {
-            const field = setArgV0.add(offset);
+const getArtMethodSpec = resolveStructSpec.bind({
```
oleavr: Did you run into a case where the new struct offset detection got the wrong offsets or failed?

iamahuman: @oleavr Yup. For Android 5.0.x 32-bit the offset calculation of `entry_point_from_quick_code_` is wrong as the width of the preceding field is always 64-bit (`uint64_t`) no matter which CPU architecture the system is on. I've had no problem for 64-bit or other versions though, so I just kept the offset detection as a fallback for future versions. It may be okay to make it a default for every versions except Lollipop, but I'm a bit unsure..

Additionally, support for 64-bit Android version is missing due to insufficient offset spec info of art::Runtime and art::ClassLinker (mandatory for correct resolution of art_quick_generic_jni_stub so that stack walking would correctly recognize it's actually a JNI method when VM is in preloaded OAT image mode). We could probably work around this by injecting a native stub method and using its quickCode as *the* JNI stub. Maybe I should work on it later...

oleavr: Ahh gotcha. That sounds good. I have a 64-bit device running 6.0, so I can have a look at figuring out those offsets. But first, let's merge this. Awesome work!

oleavr:
These improvements have now been released in Frida 8.1.14. Thanks a lot!

index.js
oleavr: @iamahuman FWIW I have now split index.js into multiple modules, and implemented calculation of all the fields:
https://github.com/frida/frida-java/blob/6f93ee08924231dae50565511d1ad1532055eedd/lib/android.js

This test is now passing with four test-devices connected -- three of which are 32-bit ARM emulators running 5.0, 5.1, and 6.0, and the last of which is a 64-bit Samsung Galaxy S6 Edge device:
https://github.com/frida/frida-java/blob/6f93ee08924231dae50565511d1ad1532055eedd/test/android.js#L22-L111
(Yeah the test is a bit verbose, but at least it's very explicit.)

Some testing still remains, but should hopefully be able to release this soon.

iamahuman:
Oh, nice work! I didn't thought the class_linker_ field would be
probe-able, hope it'll work well across versions... :D

On Nov 30, 2016 4:57 AM, "Ole André Vadla Ravnås" <notifications@github.com>
wrote:

*@oleavr* commented on this pull request.
------------------------------

In index.js <https://github.com/frida/frida-java/pull/1>:

>
-        let jniCodeOffset = -1;
-        let accessFlagsOffset = -1;
-        let remaining = 2;
-        for (let offset = 0; offset !== 64 && remaining !== 0; offset += 4) {
-            const field = setArgV0.add(offset);
+const getArtMethodSpec = resolveStructSpec.bind({

@iamahuman <https://github.com/iamahuman> FWIW I have now split index.js
into multiple modules, and implemented calculation of all the fields:
https://github.com/frida/frida-java/blob/6f93ee08924231dae50565511d1ad1
532055eedd/lib/android.js

This test is now passing with four test-devices connected -- three of which
are 32-bit ARM emulators running 5.0, 5.1, and 6.0, and the last of which
is a 64-bit Samsung Galaxy S6 Edge device:
https://github.com/frida/frida-java/blob/6f93ee08924231dae50565511d1ad1
532055eedd/test/android.js#L22-L111
(Yeah the test is a bit verbose, but at least it's very explicit.)

Some testing still remains, but should hopefully be able to release this
soon.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/frida/frida-java/pull/1>, or mute the thread

PR #12, opened 2017-02-02
Support for ART on Xposed

hedger:


lib/android.js
```diff
@@ -29,12 +29,22 @@ function getApi () {
...
+      fullAPI.isXposed = '_ZNK3art9OatHeader23IsXposedOatVersionValidEv' in exports;
+      let exp = exports[fullAPI.isXposed ? '_ZN3art6mirror6Object5CloneEPNS_6ThreadEm' :
+       '_ZN3art6mirror6Object5CloneEPNS_6ThreadE'];
+      if (exp !== undefined) {
+        fullAPI['art::mirror::Object::Clone'] = new NativeFunction(exp.address, 'pointer', 
+          fullAPI.isXposed ? ['pointer', 'pointer', 'int'] : ['pointer', 'pointer']);
+        return true;
+      }
+      return false;
     }
```
oleavr: I think it would be cleaner if we made use of the `optionals` support, by having two entries and adding both of them to `optionals`. We could then check if the regular one is missing, and plug in a shim function that just forwards to the Xposed one with the extra argument set to 0. (That way the implementation doesn't need to know about Xposed.)

hedger: I'm not sure that passing an extra argument in case of basic ART will produce correct behavior on all platforms. It may break stack or overwrite register values.

oleavr: I meant at the JS API level only, so class-factory.js doesn't need to know which of the two functions it's calling.

hedger: I see. Currently `optionals` doesn't have support for `remaining` counter, so in case if none of these exports could be found it'll still return a non-functional API object. And extending `optionals` to group and count exports by their name will result in similar, if not more, complex code.

hedger: Oh, I misread the code. I'll reimplement it with `optionals`.

hedger:
Updated. Still there's a potential problem when neither of the 2 `::Clone` versions could be found, which is VERY unlikely.

lib/class-factory.js
```diff
@@ -1054,7 +1054,8 @@ function ClassFactory (vm) {
         }
 
         const thread = api['art::Thread::CurrentFromGdb']();
-        const target = api['art::mirror::Object::Clone'](methodId, thread);
+        const target = api['art::mirror::Object::Clone'](methodId, thread);  
+
```
oleavr: Let's ditch this trailing whitespace and blank line.

lib/android.js
```diff
@@ -29,13 +29,23 @@ function getApi () {
...
+      }
     },
     variables: {
       '_ZN3art7Runtime9instance_E': function (address) {
         this.runtime_instance_ptr = address;
       }
-    }
+    },
+    optionals : ['_ZN3art6mirror6Object5CloneEPNS_6ThreadEm',
+      '_ZN3art6mirror6Object5CloneEPNS_6ThreadE']
```
oleavr: Since we're changing the data structure let's use a `Set` here instead:

```js
    optionals: new Set([
      '_ZN3art6mirror6Object5CloneEPNS_6ThreadEm',
      '_ZN3art6mirror6Object5CloneEPNS_6ThreadE',
    ]);
```

lib/android.js
```diff
@@ -101,8 +111,7 @@ function getApi () {
           }
           remaining--;
         } else {
-          const optional = optionals[name];
-          if (optional) {
+          if (optionals.indexOf(name) > -1) {
```
oleavr: Using a `Set` this becomes just:
```js
          if (optionals.has(name)) {
```

lib/android.js
```diff
@@ -29,13 +29,23 @@ function getApi () {
...
-      '_ZN3art6mirror6Object5CloneEPNS_6ThreadE': ['art::mirror::Object::Clone', 'pointer', ['pointer', 'pointer']]
+      '_ZN3art6mirror6Object5CloneEPNS_6ThreadE' : function(address) {
+        this["art::mirror::Object::Clone"] = new NativeFunction(address, 'pointer', ['pointer', 'pointer']);
+      },
+      '_ZN3art6mirror6Object5CloneEPNS_6ThreadEm' : function(address) {
+        const nativeFn = new NativeFunction(address, 'pointer', ['pointer', 'pointer', 'int']);
+        this["art::mirror::Object::Clone"] = function(thisptr, threadptr) {
+          return nativeFn(thisptr, threadptr, 0);
+        }
+      }
```
oleavr: The style here should follow [semistandard](https://github.com/Flet/semistandard) for consistency.

lib/android.js
```diff
@@ -78,7 +88,7 @@ function getApi () {
   pending.forEach(function (api) {
     const functions = api.functions || {};
     const variables = api.variables || {};
-    const optionals = api.optionals || {};
+    const optionals = api.optionals || [];
```
oleavr:
```js
    const optionals = api.optionals || new Set();
```

oleavr: Or actually let's just make `optionals` a plain array, and do the `Set` wrapping here:

```js
    const optionals = new Set(api.optionals || []);
```

hedger:
Done.

oleavr:
Thanks! Could you also revert the trailing whitespace (see comment) and tweak the style to follow [semistandard](https://github.com/Flet/semistandard)?

hedger:
Done.

lib/android.js
```diff
@@ -29,13 +29,25 @@ function getApi () {
       },
       '_ZN3art6Thread14CurrentFromGdbEv': ['art::Thread::CurrentFromGdb', 'pointer', []],
       '_ZNK3art6Thread13DecodeJObjectEP8_jobject': ['art::Thread::DecodeJObject', 'pointer', ['pointer', 'pointer']],
-      '_ZN3art6mirror6Object5CloneEPNS_6ThreadE': ['art::mirror::Object::Clone', 'pointer', ['pointer', 'pointer']]
+      '_ZN3art6mirror6Object5CloneEPNS_6ThreadE' : function(address) {
```
oleavr: - Extra space before `:`
- Missing space after `function`

lib/android.js
```diff
@@ -29,13 +29,25 @@ function getApi () {
...
+        this["art::mirror::Object::Clone"] = new NativeFunction(address, 'pointer', ['pointer', 'pointer']);
+      },
+      '_ZN3art6mirror6Object5CloneEPNS_6ThreadEm' : function(address) {
```
oleavr: - Extra space before `:`
- Missing space after `function`

lib/android.js
```diff
@@ -29,13 +29,25 @@ function getApi () {
...
+        const nativeFn = new NativeFunction(address, 'pointer', ['pointer', 'pointer', 'int']);
+        this["art::mirror::Object::Clone"] = function(thisptr, threadptr) {
+          return nativeFn(thisptr, threadptr, 0)
+        }
```
oleavr: - Missing space after `function`
- Missing semicolon inside and outside the function

lib/android.js
```diff
@@ -29,13 +29,25 @@ function getApi () {
...
+      }
     },
     variables: {
       '_ZN3art7Runtime9instance_E': function (address) {
         this.runtime_instance_ptr = address;
       }
-    }
+    },
+    optionals : new Set([
```
oleavr: - Extra space before `:`
- Let's use a plain array and do the `Set()` wrapping below (see earlier comment) – sorry for the back and forth

lib/android.js
```diff
@@ -78,7 +90,7 @@ function getApi () {
   pending.forEach(function (api) {
     const functions = api.functions || {};
     const variables = api.variables || {};
-    const optionals = api.optionals || {};
+    const optionals = api.optionals || new Set();
```
oleavr: Let's do the `Set()` wrapping here.

hedger:
Updated.

lib/android.js
```diff
@@ -29,13 +29,25 @@ function getApi () {
...
       '_ZN3art6Thread14CurrentFromGdbEv': ['art::Thread::CurrentFromGdb', 'pointer', []],
       '_ZNK3art6Thread13DecodeJObjectEP8_jobject': ['art::Thread::DecodeJObject', 'pointer', ['pointer', 'pointer']],
-      '_ZN3art6mirror6Object5CloneEPNS_6ThreadE': ['art::mirror::Object::Clone', 'pointer', ['pointer', 'pointer']]
+      '_ZN3art6mirror6Object5CloneEPNS_6ThreadE': function (address) {
+        this["art::mirror::Object::Clone"] = new NativeFunction(address, 'pointer', ['pointer', 'pointer']);
+      },
+      '_ZN3art6mirror6Object5CloneEPNS_6ThreadEm': function (address) {
+        const nativeFn = new NativeFunction(address, 'pointer', ['pointer', 'pointer', 'int']);
+        this["art::mirror::Object::Clone"] = function (thisptr, threadptr) {
+          return nativeFn(thisptr, threadptr, 0);
```
oleavr: Sorry, I missed this one:
`thisptr` => `thisPtr`
`threadptr` => `threadPtr`

hedger: No problem, renamed.

lib/android.js
```diff
@@ -29,13 +29,25 @@ function getApi () {
...
       '_ZN3art6Thread14CurrentFromGdbEv': ['art::Thread::CurrentFromGdb', 'pointer', []],
       '_ZNK3art6Thread13DecodeJObjectEP8_jobject': ['art::Thread::DecodeJObject', 'pointer', ['pointer', 'pointer']],
-      '_ZN3art6mirror6Object5CloneEPNS_6ThreadE': ['art::mirror::Object::Clone', 'pointer', ['pointer', 'pointer']]
+      '_ZN3art6mirror6Object5CloneEPNS_6ThreadE': function (address) {
+        this["art::mirror::Object::Clone"] = new NativeFunction(address, 'pointer', ['pointer', 'pointer']);
+      },
+      '_ZN3art6mirror6Object5CloneEPNS_6ThreadEm': function (address) {
+        const nativeFn = new NativeFunction(address, 'pointer', ['pointer', 'pointer', 'int']);
+        this["art::mirror::Object::Clone"] = function (thisPtr, threadPtr) {
+          return nativeFn(thisptr, threadptr, 0);
```
oleavr: Thanks!

^ these two guys also need to be renamed.

One more thing, I noticed Xposed uses `size_t` for the last argument – we don't have a type for that yet in the `NativeFunction` API, so the most correct thing would be `pointer` as it's the same size (and use the `NULL` short-hand). That will look pretty weird though, so maybe we can make that clearer by doing:

```js
          const numTargetBytes = NULL;
          return nativeFn(thisptr, threadptr, numTargetBytes);
```

hedger: Sorry, my bad. NULL trick looks reasonable, added.

lib/android.js
```diff
@@ -29,13 +29,26 @@ function getApi () {
       },
       '_ZN3art6Thread14CurrentFromGdbEv': ['art::Thread::CurrentFromGdb', 'pointer', []],
       '_ZNK3art6Thread13DecodeJObjectEP8_jobject': ['art::Thread::DecodeJObject', 'pointer', ['pointer', 'pointer']],
-      '_ZN3art6mirror6Object5CloneEPNS_6ThreadE': ['art::mirror::Object::Clone', 'pointer', ['pointer', 'pointer']]
+      '_ZN3art6mirror6Object5CloneEPNS_6ThreadE': function (address) {
+        this["art::mirror::Object::Clone"] = new NativeFunction(address, 'pointer', ['pointer', 'pointer']);
+      },
+      '_ZN3art6mirror6Object5CloneEPNS_6ThreadEm': function (address) {
+        const nativeFn = new NativeFunction(address, 'pointer', ['pointer', 'pointer', 'int']);
```
oleavr: Thanks! We'll also have to change the last argument here to `'pointer'`.

hedger: done.

oleavr:
Thanks!

PR #13, opened 2017-02-03
makefile fixes

giantpune:
if calling `make` from a makefile, use `$(MAKE)` which allows `-j` option to work properly
`adb push` doesn work like `cp` where you can `cp src1 src2 src3 dst`.  you have to push each file individually in `build/armeabi-v7a/runner build/tests.dex build/frida-java.js`

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -9,7 +9,7 @@ frida_version := 9.1.1
 test_sources := $(wildcard re/frida/*.java)
 test_classes := $(patsubst %.java,%.class,$(test_sources))
 
-android_jar := $(ANDROID_SDK_ROOT)/platforms/android-23/android.jar
+android_jar := $(shell echo $(ANDROID_SDK_ROOT))/platforms/android-23/android.jar
 
 all:
 	@echo "Available targets: deploy run"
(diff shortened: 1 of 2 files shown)


test/Makefile
```diff
@@ -9,14 +9,14 @@ frida_version := 9.1.1
...
-android_jar := $(shell echo $(ANDROID_SDK_ROOT))/platforms/android-23/android.jar
+android_jar := $(ANDROID_SDK_ROOT)/platforms/android-23/android.jar
 
 all:
 	@echo "Available targets: deploy run"
 
 deploy: build/armeabi-v7a/runner build/tests.dex build/frida-java.js
 	adb shell "rm -rf $(deploy_data_dir) && mkdir -p $(deploy_data_dir)"
-	adb push $^ $(deploy_data_dir)
+	$(foreach foo, $(^), adb push $(foo) $(deploy_data_dir))
```
oleavr: This is supported by the latest `adb` so let's be blunt and assume that :)

test/Makefile
```diff
@@ -9,7 +9,7 @@ frida_version := 9.1.1
 test_sources := $(wildcard re/frida/*.java)
 test_classes := $(patsubst %.java,%.class,$(test_sources))
 
-android_jar := $(shell echo $(ANDROID_SDK_ROOT))/platforms/android-23/android.jar
+android_jar := $(ANDROID_SDK_ROOT)/platforms/android-23/android.jar
```
oleavr: The `echo` hack was to expand the tilde in the default `ANDROID_SDK_ROOT`. Perhaps we should do: `$(abspath $(ANDROID_SDK_ROOT)` instead?

PR #19, opened 2017-02-15
Support android 4

giantpune:
* we require a writable dalvik-cache directory
* create that directory and pass it in as an env variable


oleavr:
Sweet!

PR #21, opened 2017-03-01
handle field / method name collision

### LINKED ISSUE #20: Field and Method that have same names are collinding.
I'm unable to access a static public method that have the same name as a static public field.

We might need some prefix to prevent fields and method from colliding.

olifozzy:
Allow interception when a field and a method have the same name.

note : I've added some unit tests that I wasn't able to run (no build env here)

Should fix #20 :)

### DIFF
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -669,9 +669,13 @@ function ClassFactory (vm) {
           try {
             const fieldName = invokeObjectMethodNoArgs(env.handle, field, fieldGetName);
             try {
-              const fieldjsName = env.stringFromJni(fieldName);
+              var fieldjsName = env.stringFromJni(fieldName);
               const fieldHandle = env.newGlobalRef(field);
               fieldHandles.push(fieldHandle);
+              // If we have a collided method, suffix the fieldName
+              if(jsMethods.hasOwnProperty(fieldjsName)){
+                fieldjsName = "_" + fieldjsName;
+              }
               jsFields[fieldjsName] = fieldHandle;
             } finally {
               env.deleteLocalRef(fieldName);
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -669,11 +669,10 @@ function ClassFactory (vm) {
           try {
             const fieldName = invokeObjectMethodNoArgs(env.handle, field, fieldGetName);
             try {
-              var fieldjsName = env.stringFromJni(fieldName);
+              let fieldjsName = env.stringFromJni(fieldName);
               const fieldHandle = env.newGlobalRef(field);
               fieldHandles.push(fieldHandle);
-              // If we have a collided method, suffix the fieldName
-              if(jsMethods.hasOwnProperty(fieldjsName)){
+              while (jsMethods.hasOwnProperty(fieldjsName)){
                 fieldjsName = "_" + fieldjsName;
               }
               jsFields[fieldjsName] = fieldHandle;
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -670,11 +670,13 @@ function ClassFactory (vm) {
             const fieldName = invokeObjectMethodNoArgs(env.handle, field, fieldGetName);
             try {
               let fieldjsName = env.stringFromJni(fieldName);
-              const fieldHandle = env.newGlobalRef(field);
-              fieldHandles.push(fieldHandle);
-              while (jsMethods.hasOwnProperty(fieldjsName)){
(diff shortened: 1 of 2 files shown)


olifozzy:
Should fix #20 :)

lib/class-factory.js
```diff
@@ -669,9 +669,13 @@ function ClassFactory (vm) {
           try {
             const fieldName = invokeObjectMethodNoArgs(env.handle, field, fieldGetName);
             try {
-              const fieldjsName = env.stringFromJni(fieldName);
+              var fieldjsName = env.stringFromJni(fieldName);
```
oleavr: Use `let` here to allow re-assigning – `var` is a legacy keyword from ES5.

lib/class-factory.js
```diff
@@ -669,9 +669,13 @@ function ClassFactory (vm) {
...
               const fieldHandle = env.newGlobalRef(field);
               fieldHandles.push(fieldHandle);
+              // If we have a collided method, suffix the fieldName
+              if(jsMethods.hasOwnProperty(fieldjsName)){
```
oleavr: - Comment can be removed as it's already clear what the code is doing.
- Missing space after `if`.
- Theoretically there might be a method prefixed with `_` having the same name, so using `while` here would be better.

test/re/frida/MethodTest.java
```diff
@@ -118,3 +174,17 @@ public int returnZero() {
...
+    
+    int Particle2 = 2;
+
+    int Particle() {
+        return 3;
+    }
+
+    static int Particle2(){
+        return 4;
+    }
```
oleavr: Field- and method-names should follow the coding style, i.e. `camelCase`, not `CamelCase`.

olifozzy:
Ok, I have fixed according to your feedback.

lib/class-factory.js
```diff
@@ -669,9 +669,12 @@ function ClassFactory (vm) {
           try {
             const fieldName = invokeObjectMethodNoArgs(env.handle, field, fieldGetName);
             try {
-              const fieldjsName = env.stringFromJni(fieldName);
+              let fieldjsName = env.stringFromJni(fieldName);
               const fieldHandle = env.newGlobalRef(field);
               fieldHandles.push(fieldHandle);
+              while (jsMethods.hasOwnProperty(fieldjsName)){
```
oleavr: - Let's move these 3 lines to right after the `let`, as they belong together. Then put a blank line between that and the `fieldHandle` stuff (those two lines), and then one blank line before assigning to `jsFields`. Just to make the code a bit more readable.
- Missing a space before `{`.

test/re/frida/MethodTest.java
```diff
@@ -118,3 +174,17 @@ public int returnZero() {
...
+class Collider {
+    static int particle = 1;
+    
+    int particle2 = 2;
+
+    int particle() {
+        return 3;
+    }
+
+    static int particle2(){
```
oleavr: Missing space before `{`.

test/re/frida/MethodTest.java
```diff
@@ -118,3 +174,17 @@ public int returnZero() {
         return 0;
     }
 }
+
+class Collider {
+    static int particle = 1;
+
```
oleavr: Trailing whitespace.

test/re/frida/MethodTest.java
```diff
@@ -65,6 +65,62 @@ public void genericsCanBeUsed() {
         assertEquals("Badger", script.getNextMessage());
     }
 
+
```
oleavr: There's an extra blank line here.

olifozzy:
Damn I miss stylecop ^^  Should be good now :)

lib/class-factory.js
```diff
@@ -670,11 +670,13 @@ function ClassFactory (vm) {
...
-              const fieldHandle = env.newGlobalRef(field);
-              fieldHandles.push(fieldHandle);
-              while (jsMethods.hasOwnProperty(fieldjsName)){
+              while (jsMethods.hasOwnProperty(fieldjsName)) {
                 fieldjsName = "_" + fieldjsName;
               }
+
+              const fieldHandle = env.newGlobalRef(field);
+              fieldHandles.push(fieldHandle);
+
```
oleavr: Trailing whitespace.

olifozzy:
Don't see any trailing whitespace :'(  Do you mean the extra line ?
trailing

oleavr:
The last blank line isn't blank, it's got spaces before the CRLF.

PR #54, opened 2018-01-11
memoize only when the computed value is not null fixes #53

### LINKED ISSUE #53: Java.available is always false during stage=early when using gadget in script mode
I am trying to load a script onto an android app via `wrap method`. The frida lib gets injected and is loading the script as intended but `Java.available` is always false during the initial load. 

If the frida config's on_change is set to reload and once the script is reloaded (stage=late) `Java.available ` is true and everything works fine.

I needed the script to be up and running during stage=early.

script content which is compiled using frida-compile.
```
rpc.exports = {
    init: function (stage, parameters) {
        if(Java.available) {
              require('./hooks');
        }
    }
};
```

I tried doing a "Observe(setTimeout)" on `Java.available` waiting for it to be set to true before loading my hooks...but its never set to true even though the app is loaded.

cosmosgenius:
The getApi function was getting memoize for the return value `null` as mentioned in the issue #53. This PR checks if the returned value of the compute function is null or not and if it is then doesn't memoize.

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -612,9 +612,10 @@ function memoize (compute) {
   return function (...args) {
     if (!computed) {
       value = compute(...args);
-      computed = true;
+      if(value != null) {
+        computed = true;
+      }
     }
-
     return value;
   };
 }


subho007:
@oleavr this fixes it

oleavr:
Thanks! Wasn't able to review this until now as I was refactoring surrounding code. I realized this was slightly more complicated, so I implemented it [here](4de996ca69f89f95d592eaf131527f60ce61ffd1).

PR #59, opened 2018-01-30
Forward changes to array elements through hook methods

guywithface:
Copying arrays when converting between JNI and JS stops any changes made on the JS side from propagating into the JVM. This can be a problem if the method you're hooking is responsible for filling in an array that's passed in as a parameter (and you're still delegating to that method, rather than replacing it completely.)

Proposed fix is to convert a JNI array into a JS array-like object that acts as a proxy rather than a clone. Property access to array indices access the backing JNI array; a `length` attribute rounds out the array-like interface; and the backing JNI array is stored as a property in order to convert back from JS to JNI.

### DIFF
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -2550,25 +2550,49 @@ function fromJniPrimitiveArray (arr, typeName, env, getArrayLengthFunc, getArray
   if (arr.isNull()) {
     return null;
   }
-  const result = [];
+  const result = {};
   const type = getType(typeName, true, factory);
   const length = getArrayLengthFunc.call(env, arr);
-  const cArr = getArrayElementsFunc.call(env, arr);
-  if (cArr.isNull()) {
-    throw new Error("Can't get the array elements.");
-  }
-  try {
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks!

Couple of issues:
- Defining a property per element index is potentially quite expensive, so it would be great to use a `Proxy` object instead. Feel free to use [this](https://github.com/frida/frida-objc/blob/d7dd8711850a75615bfca12e0d94be60a3d7132d/index.js#L193-L244) as boilerplate.
- Would be good to rename `backingJavaArray` to `$handle` for consistency.
- The `arr` handle is subject to use-after-free, so `$handle` should hold a global reference to it. `WeakRef.bind()` could then be used to clean it up once the JS wrapper gets GCed. (Like we do elsewhere in frida-java.)

oleavr:
I'll take a stab at implementing this – been meaning to fix this issue for a while.

oleavr:
Implemented in 5f8bde9. Releasing Frida 10.6.47 with this change. Cheers!

PR #71, opened 2018-04-25
Make Java.available non-sticky

### LINKED ISSUE #68: vm and classFactory is not initialised when using gadget in script mode.
https://github.com/frida/frida-java/blob/d89c77c5da8c20a99f7ad36a33aa474c31eb00b2/index.js#L30-L36

The issue is `tryGetApi()` returns null when the `initialize` function is called the first time, as the env is loaded by then. By the time api is ready its already too late as `initialize` is never called again.

I have tested by moving 
```
    if (api !== null) {
      vm = new VM(api);
      classFactory = new ClassFactory(vm);
    }
```
below https://github.com/frida/frida-java/blob/d89c77c5da8c20a99f7ad36a33aa474c31eb00b2/index.js#L47 and it seems to work.

I can open a PR with the same if the solution seems fine!.

cosmosgenius:
So it can be queried at multiple points during early instrumentation.

Fixes #68.

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,11 +28,7 @@ function Runtime () {
   let cachedIsAppProcess = null;
 
   function initialize () {
-    const api = tryGetApi();
-    if (api !== null) {
-      vm = new VM(api);
-      classFactory = new ClassFactory(vm);
-    }
+    tryGetApi();
   }
 
   function tryGetApi () {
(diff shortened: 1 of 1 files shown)


oleavr:
Landed in 215b108 with a few minor style tweaks. Thanks!

PR #83, opened 2018-07-25
Changed ClassLinker discovery to properly track it on Android 8.0

stjoannou:
Fix "Unable to determine ClassLinker field offsets" on Android 8.0


oleavr:
Thanks!

PR #85, opened 2018-07-26
improve the README

muhzii:
Get frida's latest release version from github api.


oleavr:
Thanks! Normally this would be great, but we don't want tests to suddenly break if a new major version of Frida is released. It should also be possible to check out an older version of frida-java and be able to run the tests.

The README changes would be great to have though!

PR #92, opened 2018-09-25
Fix return value handling for strings

muhzii:
it should also fix: #87, which happens when `GetStringUTFChars` is invoked.
[Reference](https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html#string_operations)


oleavr:
Thanks! This looks like it could really use a test to avoid future breakage – would you be interested in writing one?

oleavr:
On a second look, I think we should get this merged sooner rather than later.

oleavr:
Frida 12.2.6 is currently being released with this fix. Thanks a lot! 🙌

PR #93, opened 2018-09-26
add test case for returning strings

muhzii:
Add test method to check returning of `CharacterSequence` types (#92 ).


oleavr:
Thanks!

PR #114, opened 2019-06-01
Adding missing STD_VECTOR_SIZE definition.

eugenekolo:
sizeof(std::vector<int>) returned 24 on my 64-bit build, and 12 on 32-bit. So ti's the same as STD_STRING_SIZE, 3 pointers, begin, end, capacity.

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -16,6 +16,7 @@ const ENV_VTABLE_OFFSET_EXCEPTION_CLEAR = 17 * pointerSize;
 const ENV_VTABLE_OFFSET_FATAL_ERROR = 18 * pointerSize;
 
 const STD_STRING_SIZE = 3 * pointerSize;
+const STD_VECTOR_SIZE = 3 * pointerSize;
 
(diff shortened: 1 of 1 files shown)


eugenekolo:
I'm thinking more about this... and this assumption (also true for STD_STRING_SIZE) seems dangerous. What STL are we assuming? It may vary on devices?

oleavr:
Thanks! This got lost in the refactoring part of fdbbe1e4af203b3b22b01fe491bef71d4dd81bca.

I'm aware this is a dangerous assumption, and I suspect over time such logic will crystallise into its own module for dealing with C++/STL. We are currently assuming libc++, so this only holds true for Android >= 7, which is the point where AOSP moved from GCC to Clang.

oleavr:
Btw, the only code-path where we rely on our `std::vector` wrapper is for `Java.choose()` on Android >= 8.1, so we won't need to generalize it until we need it in code-paths used on Android < 7.

Issue #76, opened 2018-06-04
Anyway to add fields to registerClass(...)


wakawaka54:
Is there anyway to add fields to a class created through the registerClass API?

i.e.

```java
Java.registerClass({
   name: 'com.test.RunnableComponent',
   implements: [Runnable]
   fields: {
      'myField': {
           accessor: 'private',
           type: 'java.lang.String' 
      }
   },
   methods: {
       'run': function() {
           console.log(this.myField);
       }
  }
});

var test = Java.use('com.test.RunnableComponent').$init();
test.myField = 'a string';
test.run();
```

Expected Log:
**a string**

oleavr:
Not yet, but I would encourage you to add a test for it [here](https://github.com/frida/frida-java/blob/master/test/re/frida/ClassCreationTest.java) and take a stab at implementing it. What I usually do is:
1. Edit [this](https://github.com/frida/frida-java/blob/b840b14ba7ccd503eb7e794af675a1983e87abd4/test/re/frida/TestRunner.java#L23-L27) part of TestRunner.java to only run the test-suite I'm focusing on. In this case that means `"re.frida.ClassCreationTest"`.
2. Comment out the `@Test` annotations for the other tests in that suite.
3. Run `make check` once to compile the Java code and verify that the test I wrote does indeed fail as expected.
4. Run `make develop` so it monitors the JavaScript code for changes and auto-runs the test whenever there's a chance. This way I get really fast feedback as I work on the implementation.

And there's definitely a way more elegant way to do 1. and 2, in case you're interested in looking into that. :-)

Also, to improve performance and get an even faster feedback loop, edit [test/config.mk](https://github.com/frida/frida-java/blob/master/test/config.mk) and use an x86 emulator, e.g. Genymotion.

gebing:
@oleavr 
I made an implement and pull request for this feature.

PR #130, opened 2019-08-07
Enable add fields to registerClass() of Issue #76

gebing:
I made an improvement for registerClass() to enable add user defined fields, such like below:

```
var UserDefinedFields = Java.registerClass({
  name: 're.frida.SimpleFormatter',
  implements: [Formatter],
  fields: {
    fieldInt: 'int',
    fieldStr: 'java.lang.String',
  },
  methods: {
    format: [{
      returnType: 'java.lang.String',
      argumentTypes: ['int'],
      implementation: function (val) {
        const self = Java.cast(this, UserDefinedFields);
        const oldVal = self.fieldInt.value;
        self.fieldInt.value = val;
        return oldVal + ': ' + val;
      }
    }, {
      returnType: 'java.lang.String',
      argumentTypes: ['java.lang.String'],
      implementation: function (val) {
        const self = Java.cast(this, UserDefinedFields);
        const oldVal = self.fieldStr.value;
        self.fieldStr.value = val;
        return oldVal + ': ' + val;
      }
    }]
  }
});
```

### DIFF
--- a/lib/mkdex.js
+++ b/lib/mkdex.js
@@ -435,7 +435,7 @@ class DexBuilder {
 }
 
 function makeClassData (klass, constructorCodeOffset) {
-  const {constructorMethod, virtualMethods, instanceFields} = klass.classData;
+  const {instanceFields, constructorMethod, virtualMethods} = klass.classData;
 
   const staticFieldsSize = 0;
   const instanceFieldsSize = instanceFields.length;
@@ -733,6 +733,8 @@ function computeModel (classes) {
       annotationDirectories.push(annotationsDirectory);
     }
 
+    const instanceFields = fieldItems.map((field, index) => [index, kAccPublic]);
+
     const constructorNameIndex = stringToIndex['<init>'];
     const constructorMethod = classMethods
       .filter(([, name]) => name === constructorNameIndex)
@@ -752,13 +754,12 @@ function computeModel (classes) {
       .map(([index]) => {
         return [index, kAccPublic | kAccNative];
       }));
-    const instanceFields = fieldItems.map((field, index) => [index, kAccPublic]);
 
     const classData = {
+      instanceFields,
       constructorMethod,
       superConstructorMethod,
       virtualMethods,
-      instanceFields,
       offset: -1
     };
 
--- a/lib/mkdex.js
+++ b/lib/mkdex.js
@@ -11,6 +11,7 @@ const kEndianTag = 0x12345678;
 
 const kClassDefSize = 32;
 const kProtoIdSize = 12;
+const kFieldIdSize = 8;
 const kMethodIdSize = 8;
 const kTypeIdSize = 4;
 const kStringIdSize = 4;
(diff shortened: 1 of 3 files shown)


lib/mkdex.js
```diff
@@ -713,11 +752,13 @@ function computeModel (classes) {
       .map(([index]) => {
         return [index, kAccPublic | kAccNative];
       }));
+    const instanceFields = fieldItems.map((field, index) => [index, kAccPublic]);
```
oleavr: Nitpick: let's move this right before `const constructorNameIndex = `, so the order is “fields followed by methods” here as well.

lib/mkdex.js
```diff
@@ -713,11 +752,13 @@ function computeModel (classes) {
...
 
     const classData = {
       constructorMethod,
       superConstructorMethod,
       virtualMethods,
+      instanceFields,
```
oleavr: Nitpick: let's move this right before `constructorMethod`, so the order is “fields followed by methods” here as well.

test/re/frida/ClassCreationTest.java
```diff
@@ -25,6 +25,7 @@
     private static Class simplePrimitiveArrayClass = null;
     private static Class myOutputClass1 = null;
     private static Class myOutputClass2 = null;
+    private static Class userDefinedFieldClass = null;
```
oleavr: Let's move this after `orangeClass`, so these are defined in the same order as they're used.

test/re/frida/ClassCreationTest.java
```diff
@@ -310,6 +311,45 @@ public void derivedInterfaceCanBeImplemented() throws ClassNotFoundException, In
...
+    // Issue #76/#133
+    @Test
+    public void classWithUserDefinedFieldsCanBeImplemented() throws ClassNotFoundException, InstantiationException, IllegalAccessException, IOException {
+        loadScript("var Formatter = Java.use('re.frida.Formatter');" +
+                "var UserDefinedFields = Java.registerClass({" +
+                "  name: 're.frida.SimpleFormatter'," +
+                "  implements: [Formatter]," +
+                "  fields: {" +
+                "    fieldInt: 'int'," +
+                "    fieldStr: 'java.lang.String'," +
```
oleavr: How about naming these `lastInt` and `lastStr`?

test/re/frida/ClassCreationTest.java
```diff
@@ -310,6 +311,45 @@ public void derivedInterfaceCanBeImplemented() throws ClassNotFoundException, In
         assertEquals(Arrays.equals(new String[] { "tasty", "sweet" }, fruit.getTags()), true);
     }
 
+    // Issue #76/#133
+    @Test
+    public void classWithUserDefinedFieldsCanBeImplemented() throws ClassNotFoundException, InstantiationException, IllegalAccessException, IOException {
+        loadScript("var Formatter = Java.use('re.frida.Formatter');" +
+                "var UserDefinedFields = Java.registerClass({" +
+                "  name: 're.frida.SimpleFormatter'," +
```
oleavr: This name conflicts with one of the other tests. How about naming it `re.frida.StatefulFormatter` or something like that? (And renaming both variables to match.)

Issue #136, opened 2019-08-13
Memory leak of Java.registerClass's method call


gebing:
When i create a class by Java.registerClass(), and invoke method calls continuously, the jvm will terminate by global reference table overflow, the log is like blow:

```
08-12 19:32:18.072 3301-3587/? A/art: art/runtime/indirect_reference_table.cc:115] JNI ERROR (app bug): global reference table overflow (max=51200)
08-12 19:32:18.072 3301-3587/? A/art: art/runtime/indirect_reference_table.cc:115] global reference table dump:
08-12 19:32:18.072 3301-3587/? A/art: art/runtime/indirect_reference_table.cc:115]   Last 10 entries (of 51200):
08-12 19:32:18.072 3301-3587/? A/art: art/runtime/indirect_reference_table.cc:115]     51199: 0x12d4ce80 dalvik.system.PathClassLoader
08-12 19:32:18.072 3301-3587/? A/art: art/runtime/indirect_reference_table.cc:115]     51198: 0x12e9efa0 dalvik.system.DexClassLoader
08-12 19:32:18.072 3301-3587/? A/art: art/runtime/indirect_reference_table.cc:115]     51197: 0x1324da50 com.example.UserClass
08-12 19:32:18.072 3301-3587/? A/art: art/runtime/indirect_reference_table.cc:115]     51196: 0x1324da50 com.example.UserClass
08-12 19:32:18.072 3301-3587/? A/art: art/runtime/indirect_reference_table.cc:115]     51195: 0x1324da50 com.example.UserClass
08-12 19:32:18.072 3301-3587/? A/art: art/runtime/indirect_reference_table.cc:115]     51194: 0x1324da50 com.example.UserClass
08-12 19:32:18.072 3301-3587/? A/art: art/runtime/indirect_reference_table.cc:115]     51193: 0x1324da50 com.example.UserClass
08-12 19:32:18.072 3301-3587/? A/art: art/runtime/indirect_reference_table.cc:115]     51192: 0x1324da50 com.example.UserClass
08-12 19:32:18.072 3301-3587/? A/art: art/runtime/indirect_reference_table.cc:115]     51191: 0x1324da50 com.example.UserClass
08-12 19:32:18.072 3301-3587/? A/art: art/runtime/indirect_reference_table.cc:115]     51190: 0x1324da50 com.example.UserClass
08-12 19:32:18.072 3301-3587/? A/art: art/runtime/indirect_reference_table.cc:115]   Summary:
[...]

PR #137, opened 2019-08-13
Fixed for Issue #136

gebing:


gebing:
I prefer to make a method to explicitly get an extra handle/wrapper.

gebing:
Actually there is no need for a new method, just use Java.cast() which can explicitly get an extra handle/wrapper.

oleavr:
Agreed. I also thought of that, but I think `Java.cast()` sounds counter-intuitive. I think we want a function with a descriptive name; even if all it does is call `Java.cast()`. Some ideas:
- `Java.dup()`
- `Java.ref()`
- `Java.steal()` (this one would prevent the `.dispose()` from happening, which is probably a bit too clunky to implement)
- `.$ref()`
...

Food for thought, let's merge this in the meantime. :-)

gebing:
From my personal opinion, I prefer to use `Java.dup()` or `Java.ref()`.

oleavr:
Sounds good. Another idea is `Java.retain()`; what do you think?

gebing:
`Java.retain()` sounds more good.

Issue #141, opened 2019-08-27
Simultaneously invoke Java.use() will cause return wrong value


gebing:
When many threads simultaneously invoke Java.use(), It will have chance to return wrong value which does have class's properties. It is caused by the code `function ensureClass`:

```
    classes[name] = klass;

    function initializeClass () {
      ...
    }

    initializeClass();

    // Guard against use-after-"free"
    env.deleteLocalRef(classHandle);
    classHandle = null;
    env = null;

    return klass;
    
```

`initializeClass()` is called after `classes[name] = klass` and may take some time to complete. If now another thread calls `Java.use` with same class name, frida will return immediately with the value in `classes[name]` which still does not complete `initializeClass()`. It will cause error when accessing class's methods or fields.

I think `classes[name] = klass` should be put after `initializeClass()`.

PR #142, opened 2019-08-27
Fix for Issue #141

gebing:
Fix for Issue #141, Simultaneously invoke Java.use() will cause return wrong value.

### DIFF
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -194,7 +194,7 @@ function ClassFactory (vm) {
   };
 
   function getUsedClass (className) {
-    let klass;
+    let kclass;
     while ((kclass = classes[className]) === PENDING_USE) {
       Thread.sleep(0.05);
     }
@@ -204,7 +204,7 @@ function ClassFactory (vm) {
     return kclass;
   }
 
-  function setUsedClass(className, kclass) {
+  function setUsedClass (className, kclass) {
     if (kclass !== undefined) {
       classes[className] = kclass;
     } else {
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -194,7 +194,7 @@ function ClassFactory (vm) {
   };
 
   function getUsedClass(className) {
-    let kclass;
+    let klass;
     while ((kclass = classes[className]) === PENDING_USE) {
       Thread.sleep(0.05);
     }
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -145,7 +145,7 @@ function ClassFactory (vm) {
     while ((kclass = classes[className]) === PENDING_USE) {
       Thread.sleep(0.05);
     }
-    if (!kclass) {
+    if (kclass === undefined) {
       classes[className] = PENDING_USE;
     }
     return kclass;
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -78,6 +78,7 @@ function ClassFactory (vm) {
     suffix: 'dat'
   };
   const PENDING_CALLS = Symbol('PENDING_CALLS');
+  const PENDING_USE = Symbol('PENDING_USE');
 
   function initialize () {
     api = getApi();
@@ -139,8 +140,27 @@ function ClassFactory (vm) {
     }
   });
 
+  function getUsedClass(className) {
+    let kclass;
+    while ((kclass = classes[className]) === PENDING_USE) {
+      Thread.sleep(0.05);
+    }
+    if (!kclass) {
+      classes[className] = PENDING_USE;
+    }
+    return kclass;
+  }
+
+  function setUsedClass(className, kclass) {
+    if (kclass) {
+      classes[className] = kclass;
+    } else {
+      delete classes[className];
+    }
+  }
+
   this.use = function (className) {
-    let C = classes[className];
+    let C = getUsedClass(className);
     if (!C) {
       const env = vm.getEnv();
       if (loader !== null) {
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks! I think this introduces another issue, where `Java.use()` might return two different wrappers for the same class name. So I think we want to put in a "placeholder" value as early as we can, and if another thread finds this it will have to wait until the real value is there or placeholder is gone. As we don't currently have a wait primitive, we can use a loop with `Thread.sleep()` as a temporary solution.

gebing:
I think the case of `Java.use()` return two different wrappers will only rarely occur in the early stage of frida. Once `Java.use()` successfully returned, the following calls will use the wrapper from `classes[name]`. I think the only flaw is a little bit memory consumption and it will be free by garbage collection, and I think the "placeholder" solution will make the code to be a little bit complex.

gebing:
If you still want the "placeholder", I can make a implementation and update the Pull Request.

oleavr:
Yeah, I think we should ensure that it always does, to avoid surprises.

gebing:
Sorry, i do not clearly understand what you mean. Keep the current or add "placeholder"?

oleavr:
Sorry, I wasn't clear. Placeholder.

gebing:
ok

gebing:
@oleavr 
How about this pull request? When can release this patch?

lib/class-factory.js
```diff
@@ -139,8 +140,27 @@ function ClassFactory (vm) {
...
+    return kclass;
+  }
+
+  function setUsedClass(className, kclass) {
+    if (kclass) {
+      classes[className] = kclass;
+    } else {
+      delete classes[className];
+    }
+  }
```
oleavr: Let's move these below where they're used. (Code should be readable from top to bottom, higher level details at the top with details further down.)

lib/class-factory.js
```diff
@@ -139,8 +140,27 @@ function ClassFactory (vm) {
     }
   });
 
+  function getUsedClass(className) {
+    let kclass;
+    while ((kclass = classes[className]) === PENDING_USE) {
+      Thread.sleep(0.05);
+    }
+    if (!kclass) {
```
oleavr: Checks should be explicit:

```suggestion
    if (kclass === undefined) {
```

lib/class-factory.js
```diff
@@ -139,8 +140,27 @@ function ClassFactory (vm) {
...
+      classes[className] = PENDING_USE;
+    }
+    return kclass;
+  }
+
+  function setUsedClass(className, kclass) {
+    if (kclass) {
```
oleavr:
```suggestion
    if (kclass !== undefined) {
```

lib/class-factory.js
```diff
@@ -695,10 +723,7 @@ function ClassFactory (vm) {
   this.registerClass = registerClass;
 
   function ensureClass (getClassHandle, name) {
-    let klass = classes[name];
-    if (klass !== undefined) {
-      return klass;
-    }
+    let klass;
```
oleavr: This can be moved down to right before it's used.

lib/class-factory.js
```diff
@@ -715,10 +740,15 @@ function ClassFactory (vm) {
         return superHandle;
       };
 
-      try {
-        superKlass = ensureClass(getSuperClassHandle, env.getClassName(superHandle));
-      } finally {
-        env.deleteLocalRef(superHandle);
+      const superClassName = env.getClassName(superHandle);
+      superKlass = getUsedClass(superClassName);
+      if (!superKlass) {
```
oleavr:
```suggestion
      if (superKlass === undefined) {
```

lib/class-factory.js
gebing: Sorry, I can not find the right place for it. Do you mean only move `delete classes[className]` down, or move the whole function `setUsedClass()` down? 

Function  `setUsedClass()` is used by funtion `use()` and `ensureClass()`, so must be placed outside these two functions. And i also think the code of  `delete classes[className]` should not be move outside `setUsedClass()`.

Pls give me some suggestion.

oleavr: I mean the whole function. Functions can be declared below where they're used. (And ideally as close to the root closure as possible – so they capture as little as possible.)

lib/class-factory.js
```diff
@@ -177,13 +182,36 @@ function ClassFactory (vm) {
...
+        } finally {
+          setUsedClass(className, C);
+        }
       }
     }
 
     return new C(null);
   };
 
+  function getUsedClass(className) {
```
oleavr:
```suggestion
  function getUsedClass (className) {
```

Could you run this code through [semistandard](https://github.com/standard/semistandard)? Noticed missing space after method names (like this), but there might be other minor issues needing fixing.

lib/class-factory.js
```diff
@@ -177,13 +182,36 @@ function ClassFactory (vm) {
...
+    let kclass;
```
oleavr:
```suggestion
    let klass;
```

^ Perhaps we should name it `klass` for consistency?

lib/class-factory.js
```diff
@@ -177,13 +182,36 @@ function ClassFactory (vm) {
...
+      Thread.sleep(0.05);
+    }
+    if (kclass === undefined) {
+      classes[className] = PENDING_USE;
+    }
+    return kclass;
+  }
+
+  function setUsedClass(className, kclass) {
+    if (kclass !== undefined) {
```
oleavr: (Here as well.)

oleavr:
Thanks!

Issue #143, opened 2019-09-05
Java's fields should be accessible even without an instance


gebing:
Java wrapper object's fields for java class returned by `Java.use` function can not access when it is not an instance. It will throw an exception of `Unable to access instance field without an instance` by the following code of `function addMethodsAndFields ()`:

```
if (this.$handle === undefined)
  throw new Error('Unable to access instance field without an instance');
```

I think the code should be move the field value's getter/setter function, so that the field's other attributes, such as `holder`/`fieldType`/`fieldReturnType`, can be accessed when it is not an instance of class.

@oleavr 
If you agree my suggestion, i can make a pull request.

By the way, if you have time, pls check the pull request of frida/frida-gum#347. It's a bug of `hexdump`.

oleavr:
Sounds like a good idea!

Issue #139, opened 2019-08-24
Java Wrapper's getClass() should return Wrapper of 'java.lang.Class'


gebing:
Java Wrapper's `getClass()` method now return a Wrapper of `java.lang.Object`, which can not invoke `java.lang.Class`'s method, such as `getName()`/`getClassLoader()`. It must be explicitly cast to a wrapper of `java.lang.Class`. Below is the sample:

```
const clazz = Java.use('java.lang.Exception').getClass();
const classLoader = Java.cast(clazz, Java.use('java.lang.Class')).getClassLoader();
```

I digged into source code and found it was caused by below code of `Env.prototype.getTypeName` in `env.js`:

```
if (result === 'java.lang.Class' && !getGenericsInformation) {
  return this.getActualTypeArgument(type);
}
```

The code explicitly convert java.lang.Class to this actual argument type. I do not understand its purpose, but i think this was wrong.

@oleavr

gebing:
Similar case include many method such like Class.getInterfaces().

gebing:
@oleavr 

If you agree with my suggestion, i can modify the code to drop `this.getActualTypeArgument(type)`  and make a pull request for it.

oleavr:
Yes please, let's fix this! 🎉

PR #144, opened 2019-09-06
Fix for Issue #139 and Issue #143

gebing:
Fix for Issue #139, Java Wrapper's getClass() should return Wrapper of 'java.lang.Class'.
Fix for Issue #143, Java's fields should be accessible even without an instance.


oleavr:
Thank you! Sorry for the delay here, I somehow managed to miss this PR. :(

gebing:
You are welcome. :)

PR #146, opened 2019-10-10
add Runtime and ClassLinker discovery support for Android 10

Alien-AV:
Fixes "Error: Unable to determine ClassLinker field offsets" on Android 10.

The offsets at [runtime.h](https://android.googlesource.com/platform/art/+/refs/heads/android10-release/runtime/runtime.h#1000) and [class_linker.h](https://android.googlesource.com/platform/art/+/refs/heads/android10-release/runtime/class_linker.h#1369) in ART runtime changed again.

Some refactoring as well, arguably easier to read the version-dependent adjustments now.

As system apps can't be injected currently (executable-not-read permissions issue, gonna post soon), tested on non-system app.

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -410,8 +410,7 @@ function getArtVMSpec (api) {
   };
 }
 
-function _getArtRuntimeSpec (api) {
-  /*
+/*
    * class Runtime {
    * ...
    * gc::Heap* heap_;                <-- we need to find this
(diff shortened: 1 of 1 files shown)


lib/android.js
```diff
@@ -482,6 +459,37 @@ function _getArtRuntimeSpec (api) {
   return spec;
 }
 
+function getSpecByVmOffset(vm_offset, apiLevel) {
```
oleavr: Let's not refactor this in this PR, it makes future diffing very hard.

lib/android.js
```diff
@@ -621,7 +629,9 @@ function _getArtClassLinkerSpec (api) {
   for (let offset = startOffset; offset !== endOffset; offset += pointerSize) {
     const value = classLinker.add(offset).readPointer();
     if (value.equals(internTable)) {
-      const delta = (getAndroidApiLevel() >= 23) ? 3 : 5;
+      var delta = 5;
+      if(getAndroidApiLevel() >= 23) delta = 3;
```
oleavr:
```suggestion
      if (getAndroidApiLevel() >= 23) {
        delta = 3;
      }
```

This module follows [semistandard](https://github.com/standard/semistandard).

But let's move these computations outside the loop, and only call `getAndroidApiLevel()` once.

oleavr: (Actually let's not move them, it will make it harder to see what changed. Should be refactored in a separate PR.)

oleavr: We should only call `getAndroidApiLevel()` once though, since we need to change that line anyway. E.g.: `const apiLevel = getAndroidApiLevel();`

oleavr:
I'm about to cut a release so I will make the remaining tweaks so we can get this out there – this is too good to have sitting in a branch 😅

oleavr:
Tweaked version landed as d1c3e94. Thanks! 🙌

Alien-AV:
Great, thanks.

PR #147, opened 2019-12-05
Java.use()/Java.classFactory.use(): add option to bypass caching

H4oK3:
```
This patch introduces the ability to bypass the "usedClass" caching
implementation of the class-factory.js class loader. Prior to this patch, if
two separate java.lang.ClassLoader instances were to load the same class name
string, even if one were to override Java.classFactory.loader prior to each
Java.classFactory.use() call, all calls to use() after the first would result
in the cached (generated) kclass resolver function being used to load the
class, skipping the current loader entirely. For consistency, this patch also
disables calls to setUsedClass() to prevent unexpected side effects. While
it's clearly not performant to be taking the long route each time, there
appears to be no other way to deal with this issue currently as the
"usedClass" mapping is based only on the class name and not also on the
loader.
```

While doing \<REDACTED\>, me and @chaosdata noticed a situation where 
Frida was failing to hook methods of dynamically loaded classes. 
For example, the following code snippet demonstrates this issue:

```java
mBtnLoad.setOnClickListener(new View.OnClickListener() {
    @Override
    public void onClick(View view) {
        final File jarFile =
            new File("/data/local/tmp/test.dex");
        DexClassLoader dexClassLoader =
            new DexClassLoader(jarFile.getAbsolutePath(), getExternalCacheDir().getAbsolutePath(), null, getClassLoader());
        try {
            Class clazz = dexClassLoader.loadClass("com.test.Foo");
            IObject iObject = (IObject) clazz.newInstance();
            System.out.println(iObject.func());
        } catch (ClassNotFoundException e) {
            e.printStackTrace();
        } catch (InstantiationException e) {
            e.printStackTrace();
        } catch (IllegalAccessException e) {
            e.printStackTrace();
        }
    }
});
```

For the above code, we were attempting to override a method on the dynamically loaded class. But because the classloader itself is recreated, the class it loads each time is a different class. So even when overriding `Java.classFactory.loader` with the DexClassLoader instance being used each time, all attempts to use `Java.use()` on the class's name will resolve using the first DexClassLoader instance/Class that were cached. Following that, any attempts to override methods on the Frida wrapper handle value returned by `Java.use()` will result in them being applied only to the first manifested variant of the loaded Java class.

### DIFF
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -140,7 +140,8 @@ function ClassFactory (vm) {
     }
   });
 
-  this.use = function (className, skipCache) {
+  this.use = function (className, options = {}) {
+    const skipCache = options.cache === 'skip';
 
     let C = !skipCache ? getUsedClass(className) : undefined;
     if (C === undefined) {
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -140,8 +140,9 @@ function ClassFactory (vm) {
     }
   });
 
-  this.use = function (className) {
-    let C = getUsedClass(className);
+  this.use = function (className, skipCache) {
+
+    let C = !skipCache ? getUsedClass(className) : undefined;
     if (C === undefined) {
       const env = vm.getEnv();
       if (loader !== null) {
@@ -167,7 +168,9 @@ function ClassFactory (vm) {
         try {
           C = ensureClass(getClassHandle, className);
         } finally {
-          setUsedClass(className, C);
+          if (!skipCache) {
+            setUsedClass(className, C);
+          }
         }
       } else {
         const canonicalClassName = className.replace(/\./g, '/');
@@ -185,7 +188,9 @@ function ClassFactory (vm) {
         try {
           C = ensureClass(getClassHandle, className);
         } finally {
-          setUsedClass(className, C);
+          if (!skipCache) {
+            setUsedClass(className, C);
+          }
         }
       }
     }


lib/class-factory.js
```diff
@@ -140,8 +140,9 @@ function ClassFactory (vm) {
     }
   });
 
-  this.use = function (className) {
-    let C = getUsedClass(className);
+  this.use = function (className, skipCache) {
```
oleavr:
```suggestion
  this.use = function (className, options = {}) {
    const skipCache = options.cache === 'skip';
```

ChaosData: Just to clarify, is your plan to have an enum option where the value of it would be `"skip"` so that this change continues to work? Or do you plan to change the interface later to introduce it? I'm fine with this sort of thing in general though. I just did the bool/undefined truthiness check mostly to avoid having too much overhead.

oleavr: My thinking is: `Java.use('foo', { cache: 'skip' })`

I also thought of the added overhead, but considering that `Java.use()` should only be called at startup (to set up some variables at the outermost scope), and very rarely from hooks (only when it's not known or available until then), the added overhead shouldn't matter for well-written scripts.

But ISTM that the better solution here is to tie the cache to the class loader. So if you change the class loader it should also swap out the cache. But then OTOH there's another challenge: class-loaders are chained, and we don't want to re-generate the wrappers for `java.lang.*` and Android system classes. So perhaps this solution is simpler? And it solves the case where we want to get a wrapper for a class without keeping it in memory forever.

oleavr:
I'm about to cut a release so I will go ahead and make the final tweaks while landing this.

H4oK3:
Cool, looking forward to the new release :)

oleavr:
Perfect timing! Thanks :)

oleavr:
Landed as fad4bff4b8ac9897809198c6b90352ba7022606e. Thanks!

lib/class-factory.js
ChaosData: Yeah, as mentioned above, this was a quick hack to ignore the fact that the it's just the one global cache across arbitrary classloaders. My take on this is that the ideal solution would be to have multiple separate class factory instances that could be set up w/ their own classloader and then have their own cache instead of having the global class factory and swapping in and out the classloader, which is a bit race-y. I get your concerns about the potential overhead though, though I don't know enough about the inner workings of the logic to say much.

PR #148, opened 2019-12-11
Added option for: Java.use( className, { useLoaderCache: 'enable' })

H4oK3:
This patch introduces the ability to use a different class caching strategy with classLoader information. Before this patch, class wrappers are cached in `classes` objects with className as the only identifier. The new caching strategy uses `classes_with_loaders` objects to store used classes with className and classLoader object hash information. This patch also added enum options to Java.use() / Java.classFactory.use(). The new caching strategy can be turned on using `Java.use( className, { useLoaderCache: 'enable' })`

Per our conversation, .hashCode() might not be the best way to approach this; but I decided to keep it as I was not able to find a better solution to retrieve a unique identifier for a given class. Also since this patch is targeting mostly system classLoaders, there might be edge cases but it is unlikely that there are different classLoaders loading same class but with different hash code value.

### DIFF
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -65,6 +65,7 @@ function ClassFactory (vm) {
   const factory = this;
   let api = null;
   let classes = {};
+  let classes_with_loaders = {}; 
   let patchedClasses = {};
   const patchedMethods = new Set();
   const ignoredThreads = {};
@@ -108,7 +109,18 @@ function ClassFactory (vm) {
     }
 
     classes = {};
+    classes_with_loaders = {};
   };
+
+  Object.defineProperty(this, 'classes_with_loaders', {
+    enumerable: true,
+    get: function () {
+      return classes_with_loaders;
+    },
+    set: function (value) {
+      classes_with_loaders = value;
+    }
+  });
 
   Object.defineProperty(this, 'loader', {
     enumerable: true,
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks for working on this painful issue! :heart:

It seems to me that we don't want to add another option here, because the default behavior is broken to begin with and should be fixed instead.

But I wonder if it would be easier to deprecate the `loader` property and instead add something like:

```js
const factory = Java.ClassFactory.from(classLoader);
const Bar = factory.use('com.foo.Bar');
```

And call `Java.ClassFactory.from()` internally to get the factory for the app's default class loader, so existing APIs like `Java.use()` simply pass through to it.

What do you think? We could even use a Java HashMap to map each one to an `Integer` which we use to get the correct wrapper. (Whenever there's a cache miss we just increment a counter and use its value as the loader's `Integer`.)

What do you think?

oleavr:
Took a stab at it and realized it was quite straight-forward. Implemented in b6e1c33970bb2da89bd15f6312c8fd26b0ddf6cf.

Let's take it for a spin. Will push a new release now.

Cheers! :raised_hands:

H4oK3:
just saw your message; this looks good, great work!

oleavr:
Glad to hear! The only drawback that I can think of is that we now waste some CPU and memory generating class wrappers for runtime classes such as `java.lang.String`. But we can always optimize this later by looking in a shared bucket for known classes. Food for thought anyway.

PR #149, opened 2019-12-17
add getByteArrayRegion

iddoeldor:
[related issue](https://github.com/frida/frida/issues/1003)

parameters names change suggestion ? [source](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html) 

- env: the JNI interface pointer.
- array: a Java array.
- start: the starting index.
- len: the number of elements to be copied.
- buf: the destination buffer.


oleavr:
Thanks!

PR #164, opened 2020-05-29
add test for java modified utf-8

tacesrever:
as mentioned in https://github.com/frida/frida/issues/1308
in java utf-8 string 0x00 is encoded as 0xC080 instead of just 0x00, which will cause error in readUtf8String.  
see https://docs.oracle.com/javase/6/docs/api/java/io/DataInput.html#modified-utf-8

### DIFF
--- a/test/re/frida/MethodTest.java
+++ b/test/re/frida/MethodTest.java
@@ -498,6 +498,19 @@ public class MethodTest {
         assertEquals("Hello", script.getNextMessage());
     }
 
+    @Test
+    public void replacementCanAcceptModifiedUTF8StringParameter() {
+        loadScript("var Badger = Java.use('re.frida.Badger');" +
+                "Badger.eatString.implementation = function (str) {" +
+                    "send('yes');" +
+                    "return;" +
+                "};");
+
+        Badger badger = new Badger();
+        badger.feedString();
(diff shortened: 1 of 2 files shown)


test/re/frida/MethodTest.java
```diff
@@ -498,6 +498,19 @@ public void performNowWorks() {
         assertEquals("Hello", script.getNextMessage());
     }
 
+    @Test
+    public void replacementCanAcceptModifiedUTF8StringParameter() {
+        loadScript("var Badger = Java.use('re.frida.Badger');" +
+                "Badger.eatString.implementation = function (str) {" +
+                    "send('yes');" +
+                    "return;" +
```
oleavr, line 506:
```suggestion
```

test/re/frida/MethodTest.java
```diff
@@ -616,6 +629,16 @@ public String observe(String[] labels) {
         return result.toString();
     }
 
+    public void eatString(String label) {
+    }
+
+    public void feedString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("apple");
+        sb.append(String.valueOf((char)0x00));
```
oleavr, line 638:
```suggestion
        sb.append(String.valueOf((char) 0x00));
```

PR #168, opened 2020-06-14
Support for Java VM

0xra:
Hey!

I was doing some research and needed a Java instrumentation tool but Frida didn't work with JVMs yet.
Everything was pretty straight forward except for the method mangler which needed some custom code because JVMTI doesn't support the redefinition of java methods as native methods.
Enumerate classes and choose objects should work on all types of Java VMs which have JVMTI support but method interception has only been tested on HotSpot JVM on macOS (jdk >=8). Other platforms should work if the jvm library has symbols and Frida is able to find the correct APIs.

Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,6 +17,7 @@ const {
   checkJniResult
 } = require('./lib/result');
 
+const jsizeSize = 4;
 const pointerSize = Process.pointerSize;
 
 class Runtime {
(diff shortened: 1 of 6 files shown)


oleavr:
Wow, thanks so much for doing this! 🚀  Think we want to get this released as soon as possible, so I will go ahead and make some final tweaks before merging.

0xra:
That's great 🥳 Let me know if you have any questions 👍 
Next milestone should probably be early instrumentation but haven't had a need for it yet 😉

PR #170, opened 2020-06-26
Make method interception safer with VMThread

0xra:
Install new methods using VMThread::execute which blocks all Java
threads and makes it safer to do interception of hot methods.

However, compiled methods that are on the stack will still not be
flushed when calling the deoptimization logic. This is a problem
when reverting back to the old method and flushing the new method
which has been allocated using Frida’s Memory module. In this instance
we force the flushing of old compiled methods using the
NMethodSweeper::force_sweep API. We have to do it twice because of
some logic in nmethod::can_convert_to_zombie.

### DIFF
--- a/lib/jvm.js
+++ b/lib/jvm.js
@@ -367,20 +367,14 @@ class JvmMethodMangler {
     }
     this.impl = impl;
     replaceManglers.set(key, this);
-    if (!manglersScheduled) {
-      Script.nextTick(doManglers, vm);
-      manglersScheduled = true;
-    }
+    ensureManglersScheduled(vm);
   }
 
   revert (vm) {
     const { key } = this;
     replaceManglers.delete(key);
     revertManglers.set(key, this);
-    if (!manglersScheduled) {
-      Script.nextTick(doManglers, vm);
-      manglersScheduled = true;
-    }
+    ensureManglersScheduled(vm);
   }
 
   resolveTarget (wrapper, isInstanceMethod, env, api) {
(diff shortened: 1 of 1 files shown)


oleavr:
Awesome! Can we get rid of the `Thread.sleep()` calls some other way?

0xra:
It's probably safe to remove the ones before force and lower the time in withJvmThread block. The reason to have them is to make sure old compiled methods are not on the stack anymore. But then in long loops that may still not be enough time.
In java 8 we need the sleep API call so the thread can be suspended safely.

oleavr:
Makes sense! My only concern is that we're now going to make it really slow to replace 1000 methods in a row, which is very easy to do with `frida-trace -j`. So perhaps we want to expose top-level API for this, so the user can batch all of the updates. What do you think?

lib/jvm.js
```diff
@@ -348,39 +351,48 @@ class JvmMethodMangler {
...
+      this.originalMethod = mangler.originalMethod;
+      this.newMethod = mangler.newMethod;
+      this.resolved = mangler.resolved;
+    }
+    this.impl = impl;
+    replaceManglers.set(key, this);
+    if (!manglersScheduled) {
+      Script.nextTick(doManglers, vm);
+      manglersScheduled = true;
+    }
```
oleavr, line 373: Would be great to move this into a function called from here and below. E.g. `ensureManglersScheduled()` or something like that.

lib/jvm.js
```diff
@@ -395,6 +407,43 @@ class JvmMethodMangler {
...
+          force = true;
+        }
+      });
+    });
+    if (force) {
+      forceSweep(env.handle);
+    }
+    replaceManglers.clear();
+    revertManglers.clear();
+    manglersScheduled = false;
```
oleavr, line 443: I think we need to do these operations first thing, because any `NativeFunction`-call with `scheduling: 'cooperative'`, which is the default, will yield the JS lock, so another thread may in theory run JS, and start manipulating these data structures.

lib/jvm.js
0xra: Makes sense! I added the function in next commit

lib/jvm.js
0xra: I see what you're saying and I adjusted the call order in next commit

oleavr:
Thanks! Just made some style tweaks to make [semistandard](https://github.com/standard/semistandard) happy. Could you give it a quick look to make sure I didn't screw up anything? 😅

0xra:
Looks good and seems to be working just fine! 👍

PR #182, opened 2020-08-04
fixes #179 by removing kAccFastInterpreterToInterpreterInvoke from or…

### LINKED ISSUE #179: kAccFastInterpreterToInterpreterInvoke fix for missed Java hooks
Issue can be reproduced by changing `re/frida/MethodTest.java` with Pixel4 Android10:
```
@Test
public void replacementCanBeReverted() {
    Snake snake = new Snake();
    snake.die();
    loadScript("var Snake = Java.use('re.frida.Snake');" +
          "Snake.die.implementation = function () {};"); 
    snake.die();
}
...
class Snake {
    public void die() {
        System.out.println("Snake die");
    }
}
```

Expected output is _Snake die_ printed only once, however it is printed two times.

After debugging and inspecting function accessFlags it turned out that the problem is coming from ART runtime where, once function is invoked, it will be added to fast_path from  `runtime/interpreter/interpreter_common.h`. By further inspecting we can see that without previous invoke function flags are `18080001` on 2nd invoke function flags are : `58080001`. `0x40000000` is `kAccFastInterpreterToInterpreterInvoke` and matching code in `interpreter_common.h` is:

```
  // Check whether we can use the fast path. The result is cached in the ArtMethod.
  // If the bit is not set, we explicitly recheck all the conditions.
  // If any of the conditions get falsified, it is important to clear the bit.
  bool use_fast_path = false;
  if (is_mterp && self->UseMterp()) {
    use_fast_path = called_method->UseFastInterpreterToInterpreterInvoke();
    if (!use_fast_path) {
      use_fast_path = UseFastInterpreterToInterpreterInvoke(called_method);
      if (use_fast_path) {
        called_method->SetFastInterpreterToInterpreterInvokeFlag();
      }
    }
  }
```

If we clear  `kAccFastInterpreterToInterpreterInvoke`, hook will be properly invoked. Solution would be to modify `ArtMethodMangler` in `android.js` in this way:
```
    const replacementFlags = ((originalFlags & ~(kAccCriticalNative|kAccFastInterpreterToInterpreterInvoke)) | kAccNative | kAccFastNative) >>> 0;
```

deroko:
Fixes #179 thus Java methods can be hooked after those have been added to fast_path.

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -17,6 +17,7 @@ const kAccFinal = 0x0010;
 const kAccNative = 0x0100;
 const kAccFastNative = 0x00080000;
 const kAccCriticalNative = 0x00200000;
+const kAccFastInterpreterToInterpreterInvoke = 0x40000000;
 const kAccPublicApi = 0x10000000;
 const kAccXposedHookedMethod = 0x10000000;
 
@@ -1411,7 +1412,9 @@ class ArtMethodMangler {
 
     // kAccFastNative so that the VM doesn't get suspended while executing JNI
     // (so that we can modify the ArtMethod on the fly)
-    const replacementFlags = ((originalFlags & ~kAccCriticalNative) | kAccNative | kAccFastNative) >>> 0;
+    // remove kAccFastInterpreterToInterpreterInvoke to disable use_fast_path in interpreter_common.h
+    const replacementFlags = ((originalFlags & ~(kAccCriticalNative|kAccFastInterpreterToInterpreterInvoke)) | kAccNative | kAccFastNative) >>> 0;
+
 
     patchArtMethod(hookedMethod, {
       jniCode: impl,
--- a/lib/android.js
+++ b/lib/android.js
@@ -1412,7 +1412,7 @@ class ArtMethodMangler {
 
     // kAccFastNative so that the VM doesn't get suspended while executing JNI
     // (so that we can modify the ArtMethod on the fly)
-    // remove kAccFastInterpreterToInterpreterInvoke to disable use_fast_path in interpreter_common.h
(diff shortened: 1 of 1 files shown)


lib/android.js
```diff
@@ -1411,7 +1412,9 @@ class ArtMethodMangler {
 
     // kAccFastNative so that the VM doesn't get suspended while executing JNI
     // (so that we can modify the ArtMethod on the fly)
-    const replacementFlags = ((originalFlags & ~kAccCriticalNative) | kAccNative | kAccFastNative) >>> 0;
+    // remove kAccFastInterpreterToInterpreterInvoke to disable use_fast_path in interpreter_common.h
```
oleavr, line 1415:
```suggestion
    // Remove kAccFastInterpreterToInterpreterInvoke to disable use_fast_path in interpreter_common.h
```

lib/android.js
```diff
@@ -1411,7 +1412,9 @@ class ArtMethodMangler {
...
+    const replacementFlags = ((originalFlags & ~(kAccCriticalNative|kAccFastInterpreterToInterpreterInvoke)) | kAccNative | kAccFastNative) >>> 0;
```
oleavr, line 1416:
```suggestion
    const replacementFlags = ((originalFlags & ~(kAccCriticalNative | kAccFastInterpreterToInterpreterInvoke)) |
        kAccNative | kAccFastNative) >>> 0;
```

lib/android.js
```diff
@@ -1411,7 +1412,9 @@ class ArtMethodMangler {
...
+
```
oleavr, line 1417:
```suggestion
```

PR #159, opened 2020-04-01
Fix missing APIs & ClassLinker offset calculation for Android R

abdawoud:
This pull request fixes two issues related to the support of Android R (or AOSP's master builds). My setup to reproduce those issues is the following:

- x86 Cuttlefish emulator (https://source.android.com/setup/create/cuttlefish)
- System images of build # 6355543 of AOSP's master branch (https://ci.android.com/builds/branches/aosp-master/grid) for Cuttlefish emulator (aosp_cf_x86_phone)
- Testing script

> Java.perform(() => {
    console.log('Enumerating loaded classes...');
    const classes = Java.enumerateLoadedClassesSync();
    console.log(`Enumerated ${classes.length} classes`);
  });

----------

The first issue is caused by missing APIs as shown in the following error

> Error: Java API only partially available; please file a bug. Missing: _ZN3art3Dbg9StartJdwpEv, _ZN3art3Dbg8GoActiveEv, _ZN3art3Dbg21RequestDeoptimizationERKNS_21DeoptimizationRequestE, _ZN3art3Dbg20ManageDeoptimizationEv, _ZN3art3Dbg9gRegistryE
    at _getApi (../frida-java-bridge/lib/android.js:379)
    at getApi (../frida-java-bridge/lib/android.js:94)
    at _tryInitialize (../frida-java-bridge/index.js:50)
    at Runtime (../frida-java-bridge/index.js:36)
    at ../frida-java-bridge/index.js:501
    at o (node_modules/browser-pack/_prelude.js:1)
    at agent.js:1
    at /repl1.js:9557
    at o (node_modules/browser-pack/_prelude.js:1)

After fixing the first issue, the following error is thrown:

> Error: Unable to determine ClassLinker field offsets
    at _getArtClassLinkerSpec (../frida-java-bridge/lib/android.js:704)
    at ../frida-java-bridge/lib/android.js:2630
    at _getApi (../frida-java-bridge/lib/android.js:413)
    at getApi (../frida-java-bridge/lib/android.js:94)
    at _tryInitialize (../frida-java-bridge/index.js:50)
    at Runtime (../frida-java-bridge/index.js:36)
    at ../frida-java-bridge/index.js:501
    at o (node_modules/browser-pack/_prelude.js:1)
    at agent.js:1

The second error is caused by recent changes to runtime offsets in the runtime.h file (https://android.googlesource.com/platform/art/+/refs/heads/master/runtime/runtime.h#1105). See also related pull request #146.

----------

[...]

PR #185, opened 2020-09-15
Repair deoptimizeEverything() function for Android 11

Gh0u1L5:
Since Android 11, Google removed the internal JDWP implementation from ART due to the risk of bit-rot ([reference](https://android.googlesource.com/platform/art/+/fc58809f7b932d86234130be15487017dc37b0cf)). This change causes the crashes discussed in PR #159 , and breaks the deoptimizeEverything function.

After investigating the related changes carefully, I think the only solution I can suggest is "drop the JDWP deoptimization solution for Android 11 and above, and rollback to the old hacky solution based on ART internal `Instrumentation` implementations." So, yep, ¯\_(ツ)_/¯

I've tested these codes on a Pixel 3a device with latest Android 11 release, but feel free to do more tests cause there may be some other edge cases I've never know before.

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -586,27 +586,52 @@ function tryDetectInstrumentationOffset () {
 }
 
 function parsex86InstrumentationOffset (insn) {
-  if (insn.mnemonic === 'mov' && insn.operands[1].type === 'imm') {
-    return insn.operands[1].value;
+  if (insn.mnemonic !== 'mov') {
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks!  Merged with a few last-minute tweaks, most important of them being related to the `operands` property.  As each access generates the whole array with everything inside it, we should only read it once when possible.

PR #187, opened 2020-09-22
Improve the logic about detecting instrumentation offset.

Gh0u1L5:
Hi @oleavr , I've briefly checked the issue #186 , and figured out a more robust way to parse the offset for x86/x86_64 platforms. This could be helpful for future Android releases, and allow us to use `Instrumentation` on old devices in case of need.

BTW, I also added the throw back for API level >= 30, cause I thought this may be helpful for our debugging.

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -586,12 +586,15 @@ function tryDetectInstrumentationOffset () {
 }
 
 function parsex86InstrumentationOffset (insn) {
-  if (insn.mnemonic !== 'mov') {
+  if (insn.mnemonic !== 'mov' && insn.mnemonic !== 'add') {
     return null;
   }
 
   const op1 = insn.operands[1];
   if (op1.type !== 'imm') {
+    return null;
+  }
+  if (op1.value < 0x100 || op1.value > 0x400) {
     return null;
   }
 
--- a/lib/android.js
+++ b/lib/android.js
@@ -582,9 +582,6 @@ function tryDetectInstrumentationOffset () {
     cur = insn.next;
   }
 
-  if (getAndroidApiLevel() >= 30) {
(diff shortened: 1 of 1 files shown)


lib/android.js
```diff
@@ -582,18 +582,24 @@ function tryDetectInstrumentationOffset () {
     cur = insn.next;
   }
 
+  if (getAndroidApiLevel() >= 30) {
+    throw new Error('Unable to determine Runtime.instrumentation_ offset');
+  }
```
oleavr, line 587: I do like this idea, but it's probably better that we only fail when attempting to use `Java.deoptimizeEverything()`, so we're (best-case) only partially broken when running on future Android releases.

oleavr (changes requested):
Thanks a lot for looking into this! 🚀

lib/android.js
Gh0u1L5: Ah, nice catch. I agree with your opinion, and I guess line 1770 is enough for catching the problem.

lib/android.js
```diff
@@ -586,14 +586,17 @@ function tryDetectInstrumentationOffset () {
 }
 
 function parsex86InstrumentationOffset (insn) {
-  if (insn.mnemonic !== 'mov') {
+  if (insn.mnemonic !== 'mov' && insn.mnemonic !== 'add') {
```
oleavr, line 589:
```suggestion
  const { mnemonic } = insn;
  if (mnemonic !== 'mov' && mnemonic !== 'add') {
```

(Untested. Hopefully these changes will be approved by `./node_modules/.bin/semistandard`.)

oleavr (changes requested):

lib/android.js
Gh0u1L5: 👍

PR #191, opened 2020-11-26
Upgrade to using inline hooking for the ART runtime

muhzii:


lib/alloc.js
```diff
@@ -0,0 +1,40 @@
...
+  constructor (sliceSize) {
+    this.sliceSize = sliceSize;
+    this.slicesPerPage = Process.pageSize / sliceSize;
+
+    this.pages = [];
+    this.free = [];
+  }
+
+  allocateSlice () {
+    let slice = this.free.pop();
```
oleavr, line 11:
```suggestion
    const slice = this.free.pop();
```

As it's not re-assigned.

lib/alloc.js
```diff
@@ -0,0 +1,40 @@
...
+      return slice;
+    }
+
+    return this._allocatePage();
+  }
+
+  _allocatePage () {
+    const page = Memory.alloc(Process.pageSize);
+    for (let i = 1; i < this.slicesPerPage; i++) {
+      const slice = page.add(this.sliceSize * i);
```
oleavr, line 22: Would be good to avoid re-reading `this.slicesPerPage` and `this.sliceSize` here, as the JS runtime isn't necessarily smart enough to optimize these property reads.

lib/alloc.js
```diff
@@ -0,0 +1,40 @@
...
+    if (slice !== undefined) {
+      return slice;
+    }
+
+    return this._allocatePage();
+  }
+
+  _allocatePage () {
+    const page = Memory.alloc(Process.pageSize);
+    for (let i = 1; i < this.slicesPerPage; i++) {
```
oleavr, line 21: Is it intentional that the first slice isn't used? The reason we do this in Gum's code allocator is that we use that space for a header, but it doesn't look like we do that here.

oleavr, line 21:
```suggestion
    for (let i = 1; i !== this.slicesPerPage; i++) {
```

(For consistency.)

lib/android.js
```diff
@@ -1498,54 +1659,534 @@ function unwrapMethodId (methodId) {
...
+  activate (vm) {
+    this._createTrampoline();
+
+    const writeTrampoline = artQuickCodeReplacementTrampolineWriters[Process.arch];
+    const prologueLength = writeTrampoline(this.trampoline, this.quickCode, vm);
+    this.overwrittenPrologueLength = prologueLength;
+
+    const quickCodeAddress = (Process.arch === 'arm')
+      ? this.quickCode.and(THUMB_BIT_REMOVAL_MASK)
+      : this.quickCode;
```
oleavr, line 2071:
```suggestion
        ? this.quickCode.and(THUMB_BIT_REMOVAL_MASK)
        : this.quickCode;
```

Though semistandard might disagree – if it does then we'll go with what it dictates. (Did you run semistandard on this PR btw?)

lib/android.js
```diff
@@ -1498,54 +1659,534 @@ function unwrapMethodId (methodId) {
...
+    this.overwrittenPrologue = Memory.dup(quickCodeAddress, prologueLength);
+
+    const writePrologue = artQuickCodePrologueWriters[Process.arch];
+    writePrologue(this.quickCode, this.trampoline);
+  }
+
+  deactivate () {
+    const quickCodeAddress = (Process.arch === 'arm')
+      ? this.quickCode.and(THUMB_BIT_REMOVAL_MASK)
+      : this.quickCode;
```
oleavr, line 2082: Like above.

lib/android.js
```diff
@@ -1498,54 +1659,534 @@ function unwrapMethodId (methodId) {
...
+
+function isArtQuickEntrypoint (address) {
+  const api = getApi();
+
+  const m = api.module;
+
+  return address.equals(api.artClassLinker.quickGenericJniTrampoline) ||
+      address.equals(api.artClassLinker.quickToInterpreterBridgeTrampoline) ||
+      address.equals(api.artClassLinker.quickResolutionTrampoline) ||
+      address.equals(api.artClassLinker.quickImtConflictTrampoline) ||
```
oleavr, line 2106: `api.artClassLinker` is looked up repeatedly here and should ideally only be done once per call.

lib/android.js
```diff
@@ -1498,54 +1659,534 @@ function unwrapMethodId (methodId) {
...
+    });
+
+    trampolineAllocator.freeSlice(this.trampoline);
+  }
+}
+
+function isArtQuickEntrypoint (address) {
+  const api = getApi();
+
+  const m = api.module;
```
oleavr, line 2101:
```suggestion
  const { module: m, artClassLinker: linker } = api.module;
```

^ perhaps something like this.

lib/android.js
```diff
@@ -1498,54 +1659,534 @@ function unwrapMethodId (methodId) {
...
+
+    patchArtMethod(hookedMethodId, {
+      accessFlags: (originalFlags & ~(hookedMethodRemovedFlags)) >>> 0
+    }, vm);
+
+    const quickCode = this.originalMethod.quickCode;
+
+    // Replace Nterp quick entrypoints with art_quick_to_interpreter_bridge to force stepping out
+    // of ART's next-generation interpreter and use the quick stub instead.
+    if (api.artNterpEntryPoint !== undefined && quickCode.equals(api.artNterpEntryPoint)) {
```
oleavr, line 2160: Should avoid fetching the property twice.

lib/android.js
```diff
@@ -1498,54 +1659,534 @@ function unwrapMethodId (methodId) {
...
+      patchArtMethod(hookedMethodId, {
+        quickCode: api.artQuickToInterpreterBridge
+      }, vm);
+    }
+
+    if (!isArtQuickEntrypoint(quickCode)) {
+      this.interceptor = new ArtQuickCodeInterceptor(quickCode);
+      this.interceptor.activate(vm);
```
oleavr, line 2168: Nit: could be used as a local variable to avoid repeated lookups on `this`, and only assigned to `this`.

lib/android.js
```diff
@@ -1498,54 +1659,534 @@ function unwrapMethodId (methodId) {
...
+    replacedMethods.set(hookedMethodId.toString(), replacementMethodId);
+
+    notifyArtMethodHooked(hookedMethodId, vm);
   }
 
   revert (vm) {
-    patchArtMethod(this.hookedMethod, this.originalMethod, vm);
+    patchArtMethod(this.hookedMethodId, this.originalMethod, vm);
+
+    replacedMethods.delete(this.hookedMethodId.toString());
```
oleavr, line 2179: Nit: `this.hookedMethodId` is looked up twice. Could use destructuring to read it and `interceptor` in one go.

oleavr (changes requested):
Wow, this is awesome! ❤️  (Sorry it took me a while to get to this!)

lib/alloc.js
muhzii: it adds all slices to the free list except for the first one as it's being allocated for use (we could give a different name to the method if that makes it clearer).

lib/android.js
muhzii: It's applied now.

lib/alloc.js
```diff
@@ -0,0 +1,43 @@
...
+
+    return this._allocatePage();
+  }
+
+  _allocatePage () {
+    const page = Memory.alloc(Process.pageSize);
+
+    const { sliceSize, slicesPerPage } = this;
+
+    for (let i = 1; i !== slicesPerPage; i++) {
```
oleavr, line 24: Is there a reason we start at offset 1 here instead of 0? In Gum's code allocator we use the 0th slice to store metadata, but here we store that on the JS side.

oleavr (changes requested):

lib/alloc.js
oleavr: Sorry, missed your earlier comment. Please disregard this :)

lib/alloc.js
oleavr: Ahh, makes sense. Thanks for clarifying!

lib/android.js
```diff
@@ -1074,6 +1079,26 @@ function _getArtThreadSpec (vm) {
...
+        link: pointerSize
+      }
+    };
+  } else {
+    return {
+      offset: {
+        topQuickFrame: 2 * pointerSize,
+        link: 0
+      }
+    }
```
oleavr, line 1098: Missing semicolon (`semistandard` should detect this).

oleavr (changes requested):

PR #193, opened 2020-12-10
add Java.deoptimizeBootImage which calls art::Runtime::DeoptimizeBootImage

aemmitt-ns:
Adds the method `Java.deoptimizeBootImage` which is similar to deoptimizeEverything but only ensures the deoptimization of code in the boot image oat files. This method works by simply calling `art::Runtime::DeoptimizeBootImage` which uses a class visitor to set the quickCode pointer to the interpreter bridge for every class in the boot image. This is a serious performance gain in some situations where the app code itself is slow when deoptimized and it is not necessary to do in order for hooks to be hit reliably.

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -302,6 +302,7 @@ function _getApi () {
       '_ZN3art3Dbg15gDebuggerActiveE',
       '_ZN3art15instrumentation15Instrumentation20DeoptimizeEverythingEPKc',
       '_ZN3art15instrumentation15Instrumentation20DeoptimizeEverythingEv',
+      '_ZN3art7Runtime19DeoptimizeBootImageEv',
       '_ZN3art3Dbg9StartJdwpEv',
       '_ZN3art3Dbg8GoActiveEv',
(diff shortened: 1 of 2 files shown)


lib/android.js
```diff
@@ -257,6 +257,7 @@ function _getApi () {
           deoptimize(instrumentation);
         };
       },
+      _ZN3art7Runtime19DeoptimizeBootImageEv: ['art::Runtime::DeoptimizeBootImage', 'void', ['pointer']],
```
oleavr, line 260: Are we sure this is available on all ART versions? If not we should mark it as optional.

aemmitt-ns: I am checking this now. I have android 7, 10, and 11 devices to check on. Otherwise I can do emulator. But from the AOSP it looks like its there for > android 8.

oleavr: We should at least make sure it's available on 5.0, which IIRC is the oldest version where ART was default. It's also available before behind a flag IIRC, but I'm not sure we ever tested that anyway. (Oldest Android version supported is currently 4.3.)

PR #197, opened 2021-01-05
Fix error: unable to find export '_Znwm'

Happyholic1203:
When I run `frida` with openjdk-9.x on 64-bit CentOS, this happens:

```console
[Remote::PID::117235]-> Module.enumerateExports(m.path).filter(m=>/_Znwm/.test(m.name))
[]
[Remote::PID::117235]-> Module.enumerateSymbols(m.path).filter(m=>/_Znwm/.test(m.name))
[
    {
        "address": "0x2b7ea13138c0",
        "isGlobal": false,
        "name": "_Znwm",
        "section": {
            "id": "11.text",
            "protection": "r-x"
        },
        "size": 105,
        "type": "function"
    }
]
[Remote::PID::117235]-> DebugSymbol.fromName('_Znwm')
{
    "address": "0x2b7ea13138c0",
    "fileName": "",
    "lineNumber": 0,
    "moduleName": "libjvm.so",
    "name": "_Znwm"
}
```

I haven't looked into gum/gumJS to find out what caused this, but this makes frida error out whenever `Java.*` is used:

```console
Error: unable to find export '_Znwm'
```

The exact JDK I used was [9.0.4](https://download.java.net/java/GA/jdk9/9.0.4/binaries/openjdk-9.0.4_linux-x64_bin.tar.gz), which you can download from [this page](https://jdk.java.net/archive/).

I have tested the PR code using [this setup](https://gist.github.com/oleavr/cae76c895eb7d227216ed3ffe9dbbeb3) on the aforementioned JDK, and it worked.

### DIFF
--- a/lib/jvm.js
+++ b/lib/jvm.js
@@ -275,16 +275,19 @@ function _getApi () {
   }
   temporaryApi.vm = vms.readPointer();
 
-  try {
-    temporaryApi.$new = new NativeFunction(Module.getExportByName(null, '_Znwm'), 'pointer', ['ulong'], nativeFunctionOptions);
-  } catch {
-    temporaryApi.$new = new NativeFunction(ptr(DebugSymbol.fromName('_Znwm').address), 'pointer', ['ulong'], nativeFunctionOptions);
-  }
-  try {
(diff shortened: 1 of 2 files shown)


oleavr:
Thanks a lot, this is great! 🚀  Landed as d290f4e1a9b7eb1af71ff47da4436d5f468f3f34 with some minor tweaks 🙌

PR #204, opened 2021-02-09
Fix hooking of ART interpreter internals.

muhzii:
Since the replacement method that gets allocated by us is not visible to the ART runtime, it therefore
cannot be candidate for garbage collector visits when sweeps occur. In that case, the declaring_class_
field of ArtMethod (a GC root) could become stale after it gets moved to another memory space - resulting
in a crash when it is used after the fact.

Fixes https://github.com/frida/frida/issues/1609.


oleavr:
w00t, awesome! 🚀 Thanks a lot!

PR #208, opened 2021-02-19
Add ARM32 / x86 / x64 support to StackVisitor 

### LINKED ISSUE #207: Wrong signature for StackVisitor constructor
After some experimenting on a test device it looks like `_getApi()` in `android.js` cannot find the StackVisitor constructor because the referenced function name does not appear in `libart.so`.

The lookup takes place in this line:
https://github.com/frida/frida-java-bridge/blob/ae579c731ba2433889f1bdd5377b120a60876ac4/lib/android.js#L215

However, both the `libart.so` from the test device and from the official android emulator only contain these constructors:
`_ZN3art12StackVisitorC2EPNS_6ThreadEPNS_7ContextENS0_13StackWalkKindEb` 
`_ZN3art12StackVisitorC2EPNS_6ThreadEPNS_7ContextENS0_13StackWalkKindEjb`  <- Notice the "j" at the end

I suppose the latter one should be referenced. From my understanding the "j" stands for the unsigned integer / `size_t` parameter `num_frames` (see below), though I can not say for sure. Is there any documentation for how the compiler forms the function names?

```java
StackVisitor(Thread* thread,
    Context* context,
    StackWalkKind walk_kind,
    size_t num_frames,
    bool check_suspended = true)
```

P-Sc:
Fix #207 by replacing the pointer for `num_frames` with an unsigned
integer and updating the function reference in `libart.so`

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -211,6 +211,7 @@ function _getApi () {
               getInstances(instance, scope, hClass, useIsAssignableFrom, maxCount, instances);
             };
           },
+          _ZN3art12StackVisitorC2EPNS_6ThreadEPNS_7ContextENS0_13StackWalkKindEmb: ['art::StackVisitor::StackVisitor', 'void', ['pointer', 'pointer', 'pointer', 'uint', 'pointer', 'bool']],
           _ZN3art12StackVisitorC2EPNS_6ThreadEPNS_7ContextENS0_13StackWalkKindEjb: ['art::StackVisitor::StackVisitor', 'void', ['pointer', 'pointer', 'pointer', 'uint', 'uint', 'bool']],
           _ZN3art12StackVisitor9WalkStackILNS0_16CountTransitionsE0EEEvb: ['art::StackVisitor::WalkStack', 'void', ['pointer', 'bool']],
           _ZNK3art12StackVisitor9GetMethodEv: ['art::StackVisitor::GetMethod', 'pointer', ['pointer']],
--- a/lib/android.js
+++ b/lib/android.js
@@ -212,7 +212,7 @@ function _getApi () {
             };
           },
 
-          _ZN3art12StackVisitorC2EPNS_6ThreadEPNS_7ContextENS0_13StackWalkKindEmb: ['art::StackVisitor::StackVisitor', 'void', ['pointer', 'pointer', 'pointer', 'uint', 'pointer', 'bool']],
+          _ZN3art12StackVisitorC2EPNS_6ThreadEPNS_7ContextENS0_13StackWalkKindEjb: ['art::StackVisitor::StackVisitor', 'void', ['pointer', 'pointer', 'pointer', 'uint', 'uint', 'bool']],
           _ZN3art12StackVisitor9WalkStackILNS0_16CountTransitionsE0EEEvb: ['art::StackVisitor::WalkStack', 'void', ['pointer', 'bool']],
           _ZNK3art12StackVisitor9GetMethodEv: ['art::StackVisitor::GetMethod', 'pointer', ['pointer']],
           _ZNK3art12StackVisitor16DescribeLocationEv: function (address) {
@@ -1271,7 +1271,7 @@ class ArtStackVisitor {
 
     const visitor = Memory.alloc(baseSize + vtableSize);
 
-    api['art::StackVisitor::StackVisitor'](visitor, thread, context, WalkKind[walkKind], ptr(numFrames),
+    api['art::StackVisitor::StackVisitor'](visitor, thread, context, WalkKind[walkKind], numFrames,
       checkSuspended ? 1 : 0);
 
     const vtable = visitor.add(baseSize);
--- a/lib/android.js
+++ b/lib/android.js
@@ -1364,12 +1364,6 @@ class ArtMethod {
 }
 
 function makeArtQuickFrameInfoGetter (impl) {
-  if (Process.arch !== 'arm64' && Process.arch !== 'arm') {
-    return function () {
-      throw new Error('Only supported on arm64 and arm for now');
-    };
-  }
-
   return function (self) {
     const result = Memory.alloc(12);
 
@@ -1400,6 +1394,34 @@ function _getArtQuickFrameInfoGetterThunk (impl) {
       writer.putCallAddressWithArguments(impl, ['r0', 'r1']);
       writer.putPopRegs(['r0', 'lr']);
       writer.putMovRegReg('pc', 'lr');
+    });
+    return new NativeFunction(thunk, 'void', ['pointer', 'pointer'], nativeFunctionOptions);
+  } else if (Process.arch == 'ia32') {
+    const thunk = makeThunk(64, writer => {
+      writer.putMovRegRegOffsetPtr('ecx', 'esp', 4); // result
+      writer.putMovRegRegOffsetPtr('edx', 'esp', 8); // self
+      writer.putCallAddressWithArguments(impl, ['ecx', 'edx']);
+      
+      // restore callee's stack frame
+      writer.putMovRegReg('esp', 'ebp');
+      writer.putPopReg('ebp');
+
+      writer.putRet();
+    });
+    return new NativeFunction(thunk, 'void', ['pointer', 'pointer'], nativeFunctionOptions);
+  } else if (Process.arch == 'x64') {
+    const thunk = makeThunk(64, writer => {
+
+      writer.putPushReg('rdi') // preserve result buffer pointer
+      writer.putCallAddressWithArguments(impl, ['rsi']); // self
+      writer.putPopReg('rdi');
+
+      // Struct is stored by value in the rax and edx register
+      // Write struct to result buffer
+      writer.putMovRegPtrReg('rdi', 'rax');
+      writer.putMovRegOffsetPtrReg('rdi', 8, 'edx');
+
+      writer.putRet();
     });
     return new NativeFunction(thunk, 'void', ['pointer', 'pointer'], nativeFunctionOptions);
   }
--- a/lib/android.js
+++ b/lib/android.js
@@ -1397,7 +1397,7 @@ function _getArtQuickFrameInfoGetterThunk (impl) {
   } else if (Process.arch == 'arm') {
     const thunk = makeThunk(64, writer => {
       // By call convention, we pass a pointer for the result struct
-      writer.putCallAddressWithArguments(impl, ['r0']);
+      writer.putCallAddressWithArguments(impl, ['r0', 'r1']);
       writer.putPopRegs(['r0', 'lr']);
       writer.putMovRegReg('pc', 'lr');
     });
--- a/lib/android.js
+++ b/lib/android.js
@@ -213,7 +213,7 @@ function _getApi () {
           },
 
           _ZN3art12StackVisitorC2EPNS_6ThreadEPNS_7ContextENS0_13StackWalkKindEjb: ['art::StackVisitor::StackVisitor', 'void', ['pointer', 'pointer', 'pointer', 'uint', 'uint', 'bool']],
-          _ZN3art12StackVisitorC2EPNS_6ThreadEPNS_7ContextENS0_13StackWalkKindEmb: ['art::StackVisitor::StackVisitor', 'void', ['pointer', 'pointer', 'pointer', 'uint', 'pointer', 'bool']],
+          _ZN3art12StackVisitorC2EPNS_6ThreadEPNS_7ContextENS0_13StackWalkKindEmb: ['art::StackVisitor::StackVisitor', 'void', ['pointer', 'pointer', 'pointer', 'uint', 'size_t', 'bool']],
           _ZN3art12StackVisitor9WalkStackILNS0_16CountTransitionsE0EEEvb: ['art::StackVisitor::WalkStack', 'void', ['pointer', 'bool']],
           _ZNK3art12StackVisitor9GetMethodEv: ['art::StackVisitor::GetMethod', 'pointer', ['pointer']],
           _ZNK3art12StackVisitor16DescribeLocationEv: function (address) {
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks! I believe this breaks it on 64-bit, so we will need to look for both signatures instead of just one of them.

P-Sc:
It seems you are right. I inspected `libart.so` from an 64-bit lineage OS image and found the original function name there:
`_ZN3art12StackVisitorC2EPNS_6ThreadEPNS_7ContextENS0_13StackWalkKindEmb`. I could create a wrapper to check for this.

I also noticed that `getCurrentQuickFrameInfo` is not yet supported on anything other than arm64, supposedly because `_getArtQuickFrameInfoGetterThunk` uses arm64-specific registers. I might try to add arm-support, though I would need some hints for how to proceed there.
https://github.com/frida/frida-java-bridge/blob/ae579c731ba2433889f1bdd5377b120a60876ac4/lib/android.js#L1360-L1388

P-Sc:
I experimented a bit with an ARM/Thumb thunk and _accidentally_ got a somewhat working solution. The added thunk provides me with a few sensible values when [walking the stack](https://gist.github.com/oleavr/f5b6c2ed05c502e5ad53271ffdda5168).

<details>
  <summary>See the stack walking logs</summary>

```
>>> walkStack
	Visiting method 'java.lang.String com.example.hellojni.HelloJni.stringFromJNI()' at dex PC 0xffffffff (native PC 0x0)
		ArtMethod=0x783c027c
		frameInfo={"frameSizeInBytes":128,"coreSpillMask":19936,"fpSpillMask":4294901760}
	upcall
		ArtMethod=null
		frameInfo={"frameSizeInBytes":128,"coreSpillMask":19936,"fpSpillMask":4294901760}
	Visiting method 'void com.example.hellojni.HelloJni.onCreate(android.os.Bundle)' at dex PC 0x0020
		ArtMethod=0x783c0260
		frameInfo={"frameSizeInBytes":128,"coreSpillMask":19950,"fpSpillMask":65535}
	Visiting method 'void android.app.Activity.performCreate(android.os.Bundle, android.os.PersistableBundle)' at dex PC 0x000c
		ArtMethod=0x70d34e5c
		frameInfo={"frameSizeInBytes":144,"coreSpillMask":19950,"fpSpillMask":65535}
	Visiting method 'void android.app.Activity.performCreate(android.os.Bundle)' at dex PC 0x0001
		ArtMethod=0x70d34e40
		frameInfo={"frameSizeInBytes":128,"coreSpillMask":19950,"fpSpillMask":65535}
	Visiting method 'void android.app.Instrumentation.callActivityOnCreate(android.app.Activity, android.os.Bundle)' at dex PC 0x0003
		ArtMethod=0x70bd6f30
		frameInfo={"frameSizeInBytes":144,"coreSpillMask":19950,"fpSpillMask":65535}
	Visiting method 'android.app.Activity android.app.ActivityThread.performLaunchActivity(android.app.ActivityThread$ActivityClientRecord, android.content.Intent)' at dex PC 0x0141
		ArtMethod=0x70d16c20
		frameInfo={"frameSizeInBytes":144,"coreSpillMask":19950,"fpSpillMask":65535}
	Visiting method 'android.app.Activity android.app.ActivityThread.handleLaunchActivity(android.app.ActivityThread$ActivityClientRecord, android.app.servertransaction.PendingTransactionActions, android.content.Intent)' at dex PC 0x0024
		ArtMethod=0x70d171ec
		frameInfo={"frameSizeInBytes":144,"coreSpillMask":19950,"fpSpillMask":65535}
	Visiting method 'void android.app.servertransaction.LaunchActivityItem.execute(android.app.ClientTransactionHandler, android.os.IBinder, android.app.servertransaction.PendingTransactionActions)' at dex PC 0x0038
		ArtMethod=0x70ca7d68
		frameInfo={"frameSizeInBytes":144,"coreSpillMask":19950,"fpSpillMask":65535}
	Visiting method 'void android.app.servertransaction.TransactionExecutor.executeCallbacks(android.app.servertransaction.ClientTransaction)' at dex PC 0x0062
		ArtMethod=0x70b68ffc
		frameInfo={"frameSizeInBytes":128,"coreSpillMask":19950,"fpSpillMask":65535}
	Visiting method 'void android.app.servertransaction.TransactionExecutor.execute(android.app.servertransaction.ClientTransaction)' at dex PC 0x0022
		ArtMethod=0x70b68fe0
		frameInfo={"frameSizeInBytes":128,"coreSpillMask":19950,"fpSpillMask":65535}
	Visiting method 'void android.app.ActivityThread$H.handleMessage(android.os.Message)' at dex PC 0x0027
		ArtMethod=0x70b33ae8
		frameInfo={"frameSizeInBytes":128,"coreSpillMask":19950,"fpSpillMask":65535}
	Visiting method 'void android.os.Handler.dispatchMessage(android.os.Message)' at dex PC 0x0015
		ArtMethod=0x70c367c8
		frameInfo={"frameSizeInBytes":128,"coreSpillMask":19950,"fpSpillMask":65535}
	Visiting method 'void android.os.Looper.loop()' at dex PC 0x00cf
		ArtMethod=0x70c54df0
		frameInfo={"frameSizeInBytes":128,"coreSpillMask":19950,"fpSpillMask":65535}
	Visiting method 'void android.app.ActivityThread.main(java.lang.String[])' at dex PC 0x006c
		ArtMethod=0x70d16b40
		frameInfo={"frameSizeInBytes":128,"coreSpillMask":19950,"fpSpillMask":65535}
	upcall
		ArtMethod=null
Process crashed: Bad access due to invalid address

***
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
LineageOS Version: '16.0-20190701-UNOFFICIAL-osprey'
Build fingerprint: 'motorola/osprey_retus/osprey_umts:6.0/MPI24.65-25.1/1:user/release-keys'
Revision: '0'
ABI: 'arm'
pid: 18875, tid: 18875, name: xample.hellojni  >>> com.example.hellojni <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x4
Cause: null pointer dereference
    r0  adb348a8  r1  00000000  r2  5089e1cb  r3  ae611200
    r4  aaa3c1d0  r5  00000000  r6  adbfe400  r7  bea3a4d0
    r8  ae063948  r9  bea3a450  r10 bea3a500  r11 bea3a460
    ip  00000000  sp  bea3a440  lr  ada21c89  pc  ada21c92

backtrace:
    #00 pc 00361c92  /system/lib/libart.so (offset 0x96000) (art::StackVisitor::GetCurrentQuickFrameInfo() const+38)
    #01 pc 00000005  <anonymous:ae6b7000>
***
[MotoG3::com.example.hellojni]->

Thank you for using Frida!
```
</details>

However, I'm not sure why it fails in the end and I definitely cannot tell why it works at all. This thunk

- apparently doesn't need to push the link register (LR) beforehand (as opposed to the ARM64 thunk),
- even needs to pop a another value from the stack to work,
- doesn't need the "this" pointer passed to the called function

P-Sc:
Okay I've added x86 and x64 support and everything should work now for all architectures. I cannot test it for Arm64 though, as I do not own an appropriate testing device and I cannot find a working emulator.

Small note: The point 3 of my last comment is not exactly true. It only worked because the "this" pointer already was in the right register. Still, I'm at loss for why the link register needs to be popped from the stack. By [calling convention](https://developer.arm.com/documentation/ihi0042/j/?lang=en#subroutine-calls), the link register should already hold the correct return address.

lib/android.js
```diff
@@ -211,8 +211,8 @@ function _getApi () {
               getInstances(instance, scope, hClass, useIsAssignableFrom, maxCount, instances);
             };
           },
-
```
oleavr, line 214: This blank line is intentional and should be kept for readability.

lib/android.js
```diff
@@ -211,8 +211,8 @@ function _getApi () {
...
           _ZN3art12StackVisitorC2EPNS_6ThreadEPNS_7ContextENS0_13StackWalkKindEmb: ['art::StackVisitor::StackVisitor', 'void', ['pointer', 'pointer', 'pointer', 'uint', 'pointer', 'bool']],
+          _ZN3art12StackVisitorC2EPNS_6ThreadEPNS_7ContextENS0_13StackWalkKindEjb: ['art::StackVisitor::StackVisitor', 'void', ['pointer', 'pointer', 'pointer', 'uint', 'uint', 'bool']],
```
oleavr, line 215: I suppose we should move this line above the other so these two are alphabetically sorted relative to each other.

lib/android.js
```diff
@@ -307,6 +307,7 @@ function _getApi () {
           '_ZN3art2gc4Heap12GetInstancesERNS_24VariableSizedHandleScopeENS_6HandleINS_6mirror5ClassEEEiRNSt3__16vectorINS4_INS5_6ObjectEEENS8_9allocatorISB_EEEE',
           '_ZN3art2gc4Heap12GetInstancesERNS_24VariableSizedHandleScopeENS_6HandleINS_6mirror5ClassEEEbiRNSt3__16vectorINS4_INS5_6ObjectEEENS8_9allocatorISB_EEEE',
           '_ZN3art12StackVisitorC2EPNS_6ThreadEPNS_7ContextENS0_13StackWalkKindEmb',
+          '_ZN3art12StackVisitorC2EPNS_6ThreadEPNS_7ContextENS0_13StackWalkKindEjb',
```
oleavr, line 310: (And here as well.)

lib/android.js
```diff
@@ -1271,8 +1272,13 @@ class ArtStackVisitor {
 
     const visitor = Memory.alloc(baseSize + vtableSize);
 
-    api['art::StackVisitor::StackVisitor'](visitor, thread, context, WalkKind[walkKind], ptr(numFrames),
-      checkSuspended ? 1 : 0);
+    if (pointerSize == 4) {
```
oleavr, line 1275:
```suggestion
    if (pointerSize === 4) {
```

Should always use strict equality checks.

lib/android.js
```diff
@@ -1378,16 +1378,53 @@ function makeArtQuickFrameInfoGetter (impl) {
...
-  return new NativeFunction(thunk, 'void', ['pointer', 'pointer'], nativeFunctionOptions);
+      // Struct is stored by value in the rax and edx register
+      // Write struct to result buffer
+      writer.putMovRegPtrReg('rdi', 'rax');
+      writer.putMovRegOffsetPtrReg('rdi', 8, 'edx');
+
+      writer.putRet();
+    });
+    return new NativeFunction(thunk, 'void', ['pointer', 'pointer'], nativeFunctionOptions);
+  }
```
oleavr, line 1427: - All of these thunks are 64 bytes large – can we reduce some of them? (Previously I've looked at how much code is generated, and rounded up to the nearest power-of-two.)
- Would be good to use `switch` to avoid repeatedly reading `Process.arch`
- Seems the `NativeFunction` construction could be moved out of the architecture-specific code
- For consistency the architectures should be in the following order: ia32, x64, arm, arm64

lib/android.js
```diff
@@ -1378,16 +1378,53 @@ function makeArtQuickFrameInfoGetter (impl) {
...
+      writer.putMovRegReg('esp', 'ebp');
+      writer.putPopReg('ebp');
+
+      writer.putRet();
+    });
+    return new NativeFunction(thunk, 'void', ['pointer', 'pointer'], nativeFunctionOptions);
+  } else if (Process.arch == 'x64') {
+    const thunk = makeThunk(64, writer => {
+
+      writer.putPushReg('rdi') // preserve result buffer pointer
```
oleavr, line 1415: Something is off with the indentation here. Did you run this PR through `./node_modules/.bin/semistandard`?

lib/android.js
```diff
@@ -1378,16 +1378,53 @@ function makeArtQuickFrameInfoGetter (impl) {
...
+      writer.putCallAddressWithArguments(impl, ['x1']);
+      writer.putPopRegReg('x2', 'lr');
+      writer.putStrRegRegOffset('x0', 'x2', 0);
+      writer.putStrRegRegOffset('w1', 'x2', 8);
+      writer.putRet();
+    });
+    return new NativeFunction(thunk, 'void', ['pointer', 'pointer'], nativeFunctionOptions);
+  } else if (Process.arch == 'arm') {
+    const thunk = makeThunk(64, writer => {
+      // By call convention, we pass a pointer for the result struct
```
oleavr, line 1393:
```suggestion
      // By calling convention, we pass a pointer for the result struct
```

lib/android.js
```diff
@@ -1378,16 +1378,53 @@ function makeArtQuickFrameInfoGetter (impl) {
...
+      writer.putMovRegReg('pc', 'lr');
+    });
+    return new NativeFunction(thunk, 'void', ['pointer', 'pointer'], nativeFunctionOptions);
+  } else if (Process.arch == 'ia32') {
+    const thunk = makeThunk(64, writer => {
+      writer.putMovRegRegOffsetPtr('ecx', 'esp', 4); // result
+      writer.putMovRegRegOffsetPtr('edx', 'esp', 8); // self
+      writer.putCallAddressWithArguments(impl, ['ecx', 'edx']);
+      
+      // restore callee's stack frame
```
oleavr, line 1405:
```suggestion
      // Restore callee's stack frame
```

lib/android.js
```diff
@@ -1378,16 +1378,53 @@ function makeArtQuickFrameInfoGetter (impl) {
...
+      writer.putMovRegReg('esp', 'ebp');
+      writer.putPopReg('ebp');
+
+      writer.putRet();
+    });
+    return new NativeFunction(thunk, 'void', ['pointer', 'pointer'], nativeFunctionOptions);
+  } else if (Process.arch == 'x64') {
+    const thunk = makeThunk(64, writer => {
+
```
oleavr, line 1414:
```suggestion
```

oleavr (changes requested):
Thanks! 💥

lib/android.js
P-Sc: I reduced the thunk size now. I suppose the nearest (rounded up) power-of-two is still preferable?

lib/android.js
P-Sc: I didn't know `semistandard` beforehand. In that line it only reported a missing semicolon, though.

Anyway, `semistandard` reports no problems now.

oleavr:
Thanks a lot! (Apologies for the delay here, I was stuck in another part of the codebase for much longer than I'd hoped.)

P-Sc:
No worries, happy to help :slightly_smiling_face:

PR #210, opened 2021-02-25
Fix result of Java.enumerateMethods on ART

muhzii:
This is a bug where static initializer methods are being included in the enumerated set as '$init' while they should be skipped altogether.

### DIFF
--- a/lib/class-model.js
+++ b/lib/class-model.js
@@ -599,6 +599,12 @@ collect_matching_class_methods (ArtClassVisitor * self,
     }
     else
     {
+      const gchar * name_begin;
+
+      name_begin = bare_method_name + class_name_length + 1;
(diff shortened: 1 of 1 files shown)


lib/class-model.js
```diff
@@ -599,6 +599,12 @@ collect_matching_class_methods (ArtClassVisitor * self,
     }
     else
     {
+      const gchar * name_begin;
+
+      name_begin = bare_method_name + class_name_length + 1;
+      if (is_constructor && g_str_has_prefix (name_begin, "<clinit>"))
```
oleavr, line 605: Does the string contain anything after `<clinit>`?

oleavr (commented):
Thanks!

lib/class-model.js
muhzii: Nope.

PR #211, opened 2021-03-02
Fix handling of generic array types

muhzii:
This allows array objects obtained from the runtime to be reused later when marshaling array types, which is necessary in order to preserve type information in cases where the type is dynamic.

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -429,89 +429,69 @@ function getArrayType (typeName, unbox, factory) {
       });
     },
     fromJni (arr, env, owned) {
-      // We'll ignore the owned hint as we might otherwise run out of local references.
-      return fromJniObjectArray(arr, env, function (self, elem) {
-        return elementType.fromJni(elem, env);
-      }, owned);
+      if (arr.isNull()) {
+        return null;
+      }
+
+      const result = [];
+
+      const n = env.getArrayLength(arr);
+      for (let i = 0; i !== n; i++) {
+        const element = env.getObjectArrayElement(arr, i);
+        try {
+          // We'll ignore the owned hint as we might otherwise run out of local references.
+          result.push(elementType.fromJni(element, env));
+        } finally {
+          env.deleteLocalRef(element);
+        }
+      }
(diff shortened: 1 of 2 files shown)


test/re/frida/MethodTest.java
```diff
@@ -10,6 +10,8 @@
 
 import java.io.IOException;
 import java.lang.UnsupportedOperationException;
+import java.lang.ClassCastException;
+import java.lang.reflect.Array;
```
oleavr, line 14: Imports should be kept in alphabetic order.

test/re/frida/MethodTest.java
```diff
@@ -519,6 +521,21 @@ public void nativeMethodCanBeReplaced() {
         assertEquals("randomString", script.getNextMessage());
     }
 
+    @Test
+    public void genericArrayTypeShouldBePreserved() {
+        loadScript("var GenericArray = Java.use('re.frida.GenericArray');" +
+                "var getArray2 = GenericArray.getArray2;" +
```
oleavr, line 527: Let's avoid `var` for new tests. (Would be good to port our existing tests to modern JavaScript someday.)

lib/types.js
```diff
@@ -517,6 +510,14 @@ function disposeObjectArray () {
     }
     dispose.call(obj);
   }
+
+  const wrapper = this.$w;
+  if (wrapper !== undefined) {
+    const dispose = wrapper.$dispose;
+    if (dispose !== undefined) {
+      dispose.call(wrapper);
```
oleavr, line 518: I suppose we could use nullish coalescing syntax to simplify this?

oleavr (changes requested):
Thanks! 💥

lib/types.js
muhzii: Yes, fixed. Also, I think using `call` here is not necessary? Personally, I did it that way for it to become the same as other calls to the same function. However, automatic binding should already take care of passing the correct `this` context in this case.

PR #214, opened 2021-04-02
Add chooseSync method

gergesh:


index.js
```diff
@@ -465,6 +465,18 @@ class Runtime {
...
+    const instances = [];
+    this.choose(specifier, {
+      onMatch (c) {
+        instances.push(c);
+      },
+      onComplete () {
+      }
+    });
+    return instances;
+  }
```
oleavr, line 478: This assumes that the underlying implementation is always synchronous. It unfortunately isn't in the case of Dalvik, and we may choose to make the others asynchronous later. So it would be better to return a `Promise` in `choose()` so make it easy to use with `await`.

oleavr:
Thanks for taking a stab at this nevertheless! ❤️

PR #217, opened 2021-04-15
Remove kAccSingleImplementation flag to avoid method being devirtualized

liuyufei:
Remove kAccSingleImplementation flag to avoid method being devirtualized so that frida can hook the method correctly

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -21,6 +21,7 @@ const kAccFastNative = 0x00080000;
 const kAccCriticalNative = 0x00200000;
 const kAccFastInterpreterToInterpreterInvoke = 0x40000000;
 const kAccSkipAccessChecks = 0x00080000;
+const kAccSingleImplementation = 0x08000000;
 const kAccPublicApi = 0x10000000;
 const kAccXposedHookedMethod = 0x10000000;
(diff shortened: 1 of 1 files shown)


lib/android.js
```diff
@@ -2316,6 +2317,7 @@ class ArtMethodMangler {
     // Remove kAccFastInterpreterToInterpreterInvoke and kAccSkipAccessChecks to disable use_fast_path
     // in interpreter_common.h
     let hookedMethodRemovedFlags = kAccFastInterpreterToInterpreterInvoke;
+    hookedMethodRemovedFlags |= kAccSingleImplementation;
```
oleavr, line 2320:
```suggestion
    let hookedMethodRemovedFlags = kAccFastInterpreterToInterpreterInvoke | kAccSingleImplementation;
```

oleavr (changes requested):
Thanks! 🎉

lib/android.js
oleavr: (Would be good to run `node_modules/.bin/semistandard` after this change – not sure if I exceeded the line length in this suggestion here.)

PR #221, opened 2021-05-22
Fix Java.array for unloaded array types

yotamN:
If the type of the array was not used in the application it means that it wasn't loaded in memory and Java.array failed because of this. This fix tries first to use the array type and if it fail, it will first create an empty array of that type using java.lang.reflect.Array.newInstance before calling factory.use.

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -446,7 +446,15 @@ function getArrayType (typeName, unbox, factory) {
         }
       }
 
-      result.$w = factory.cast(arr, factory.use(typeName), owned);
+      // The type name we get is not always the correct representation of the type so we make it so here.
+      const internalTypeName = '[L' + elementTypeName.replace(/\./g, '/') + ';';
+      try {
+        result.$w = factory.cast(arr, factory.use(internalTypeName), owned);
+      } catch (e) {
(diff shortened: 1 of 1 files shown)


lib/types.js
```diff
@@ -446,7 +446,15 @@ function getArrayType (typeName, unbox, factory) {
         }
       }
 
-      result.$w = factory.cast(arr, factory.use(typeName), owned);
+      // The type name we get is not always the correct representation of the type so we make it here
+      const internalTypeName = "[L" + elementTypeName.replace(/\./g, '/') + ";";
```
oleavr, line 450: For consistency:

```suggestion
      const internalTypeName = '[L' + elementTypeName.replace(/\./g, '/') + ';';
```

(Here and below.)

Could you also check that this passes semistandard? (By running `./node_modules/.bin/semistandard`.)

oleavr (changes requested):
Thanks! 😍

lib/types.js
yotamN: Fixed and rebased, sorry for the delay

oleavr: No worries. Thanks a lot for fixing it! 🙌

PR #225, opened 2021-06-15
Add toString method to primitive arrays

yotamN:
Instead of printing the generic `[Object Object]` string it would print the array value separated by commas.


oleavr:
Thanks! 💥

PR #241, opened 2022-02-03
Always prepend underscores to field names

### LINKED ISSUE #240: Always prepend an underscore for fields
Because of the way of how Frida resolve conflicts between methods and field with the same name it require the user to know if there is a method with the same name as the requested field before accessing it. I propose adding an underscore anyway, so people could always prepend an underscore to the field and expect it to work.

yotamN:
Because of the way Frida resolve name conflicts, it can be hard to determine the name of a field in Frida. To prevent this, this patch make the naming of fields consistent (as long as there isn't another conflict).

Resolve #240

### DIFF
--- a/lib/class-model.js
+++ b/lib/class-model.js
@@ -409,6 +409,7 @@ model_add_field (Model * self,
 {
   GHashTable * members = self->members;
   gchar * key, type;
+  int prepended_underscore = FALSE;
 
   key = g_strdup (name);
   while (g_hash_table_contains (members, key))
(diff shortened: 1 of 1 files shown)


oleavr:
I'm a bit on the fence on this one, since there may also be a clashing method name prefixed with an underscore. My gut feeling is that we should come up with a better approach on the next major bump. Perhaps fields should be prefixed with a character that's illegal in method names. Would also like to remove the wrapper that has the `value` property, as it's very confusing. (And the introspection has limited value anyway.)

yotamN:
>My gut feeling is that we should come up with a better approach on the next major bump

Is there such a thing? I think in both cases (Java and JavaScript) methods and properties have the same restrictions of an identifier.

It might be a bit verbose but why not expose it through two different properties? You could do it like so:

```
cls.fields.abc
cls.methods.abc()
```

or if you want it to be a bit less verbose:

```
cls.f.abc
cls.m.abc()
```

s1341:
Whatever we do we need to be careful not to break peoples existing scripts and infrastructure too badly.

oleavr:
Hmm. This will still be confusing if a method is named `_x`, since a field named `x` won't also be visible as `_x`.

yotamN:
I think I will close this PR until we decide on a solution

PR #243, opened 2022-02-16
Fix "$" method name is not handled correctly

eybisi:
If "$" is used as a method name get function of wrapper does not handle it correctly.  This fixes the issue. I also added test for this. Not sure it is necessary

For the record malware packers use "$" as string decryption function name to avoid being hooked by frida :)
dollarsign

### DIFF
--- a/lib/class-model.js
+++ b/lib/class-model.js
@@ -390,7 +390,10 @@ model_add_method (Model * self,
   gchar * key, type;
   const gchar * value;
 
-  key = g_strdup (name);
+  if (name[0] == '$')
+    key = g_strdup_printf ("_%s", name);
+  else
+    key = g_strdup (name);
 
   type = (modifiers & kAccStatic) != 0 ? 's' : 'i';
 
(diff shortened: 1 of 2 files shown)


oleavr:
Thanks! I guess we also need to handle names that are `$something`? If so they might mask our own, and that means we should pick a different prefix than `$`. We could use JS symbols for this to make it impossible for collisions to happen.

eybisi:
Correct, function names like $test() $init() also fails. Somehow we need to change $init and $new too, that would break lots of script.

oleavr:
I suppose one way to deal with it is to transform fields/methods starting with `$` by prepending an underscore (and keep doing that if there's a conflict).

CyberSaxosTiGER:
> I suppose one way to deal with it is to transform fields/methods starting with `$` by prepending an underscore (and keep doing that if there's a conflict).

we should type that into the docs. like you should use "$_" if your variable is starting with "$".

and there is already a proto variable named `$_s`. that variable could be renamed as `$ss`.

https://github.com/frida/frida-java-bridge/blob/3c8c91e82901437098d33ffb3981be1c3ae2c5ad/lib/class-factory.js#L924

eybisi:
Another option will be:
We add _ if there is a collision with frida props. Because adding _ always sound counter intuitive (I'm thinking user experience here), since hooking $example would require to write $_example, while it is not in frida props. If there is a collision and user hooks, we can inform user and say you should write $_ownProperty() instead of $ownProperty(). 

And lets say we add _ always, user hooks $example. In that case we should inform user again since it is not correct. Which path we should take ?

oleavr:
(Sorry for the delay here!) Let's go for simply underscore-prefixing methods/fields starting with `$anything` so they become `_$anything`.

PR #244, opened 2022-04-01
frida-trace: improve error message for Java method searching

jpstotz:
As shown by [#1644](https://github.com/frida/frida/issues/1644) the error message `Error: Invalid method query` is not really helpful as it does not give any hint how to change the method query parameter to make frida-trace accept it. 

I noticed that the pattern format is described in the [Java.enumerateMethods](https://frida.re/docs/javascript-api/#java) documentation so in my opinion we should change the error message to contain the basic pattern plus a hint where to find more details.

### DIFF
--- a/lib/class-model.js
+++ b/lib/class-model.js
@@ -1176,7 +1176,7 @@ class Model {
 
     const params = query.match(methodQueryPattern);
     if (params === null) {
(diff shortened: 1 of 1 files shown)


oleavr (approved):
Thanks!

Issue #262, opened 2022-08-27
Android 13: getPackageInfoNoCheck overload missing


Bleuzen:
Frida currently fails on Android 13. When starting a script, for example:

```
// frida.js

Java.perform(function() {
})
```

Command:
`frida -U -l frida.js --no-pause -f <package>`

It fails with:

```
Error: getPackageInfoNoCheck(): has more than one overload, use .overload(<signature>) to choose from:
        .overload('android.content.pm.ApplicationInfo', 'android.content.res.CompatibilityInfo')
        .overload('android.content.pm.ApplicationInfo', 'android.content.res.CompatibilityInfo', 'boolean')
    at X (frida/node_modules/frida-java-bridge/lib/class-factory.js:569)
    at K (frida/node_modules/frida-java-bridge/lib/class-factory.js:564)
    at set (frida/node_modules/frida-java-bridge/lib/class-factory.js:932)
    at <anonymous> (frida/node_modules/frida-java-bridge/index.js:224)
    at <anonymous> (frida/node_modules/frida-java-bridge/lib/vm.js:12)
    at _performPendingVmOpsWhenReady (frida/node_modules/frida-java-bridge/index.js:244)
    at perform (frida/node_modules/frida-java-bridge/index.js:204)
    at <anonymous> (/frida/repl-2.js:2)
```

---

Android version: 13
Frida server version: 15.2.2
Frida cli version: 15.2.2

---

On Android 13 there is an overload for this method now:
![image](https://user-images.githubusercontent.com/12885163/187029578-6ebf2686-4fd9-46a3-9245-985151313fc3.png)

boehs:
@oleavr when will this make it to a downloadable version of frida?

PR #263, opened 2022-08-27
fix: android 13 compatibility with `getPackageInfoNoCheck`

Bleuzen:
UNTESTED!

*Should* fix https://github.com/frida/frida-java-bridge/issues/262

Please test before merging!
I haven't yet since I currently don't know how to build this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -440,7 +440,7 @@ class Runtime {
         handleBindApplication.apply(this, arguments);
       };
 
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks for looking into this! Just landed a slightly different approach in 3f8fc17727458f8319c7d4261df27e2e401fd427.

PR #264, opened 2022-09-13
[Fix] workaround for ClassLinker offset on Android 11 & 12

sh4dowb:
Some devices are getting "Unable to determine ClassLinker field offsets".
https://github.com/frida/frida/issues/2176

This could be related to ROM, or apparently "Google Play System Update".
Changing `classLinkerOffset = offset - (3 * pointerSize);` to `classLinkerOffset = offset - (4 * pointerSize);` fixed this issue for these devices, however so far it's unclear what is causing this change.

This workaround tries with `3` first, and if it fails, it tries `4`. I had to disable memoize because for some reason it wouldn't run with new arguments. The code isn't perfect, but it works, of course improvements by the community are most welcome.

Tested the same `frida-server` on multiple devices with and without the issue, and it seems to be resolved with this patch.

If allowed, I'd like to publish binaries for android-arm until an official release is out.

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -81,7 +81,6 @@ const SOCK_STREAM = 1;
 
 const getArtRuntimeSpec = memoize(_getArtRuntimeSpec);
 const getArtInstrumentationSpec = memoize(_getArtInstrumentationSpec);
-const getArtClassLinkerSpec = memoize(_getArtClassLinkerSpec);
 const getArtMethodSpec = memoize(_getArtMethodSpec);
 const getArtThreadSpec = memoize(_getArtThreadSpec);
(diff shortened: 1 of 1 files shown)


sh4dowb:
just a note, the same build, while Frida server works on the device with said problem, patching an apk with frida-gadget still seems to cause issues. if I remember correctly, something like threads are not equal. while I doubt it, maybe it's an error related to my compilation settings.

oleavr:
Yay, thanks a lot for doing this! :raised_hands:

WindySha:
This maybe a better way to ge the offset, not need to guess it's 3 or 4: 
https://github.com/PAGalaxyLab/YAHFA/pull/173

PR #267, opened 2022-10-14
Update tests to new major version 16

yotamN:
There wasn't much to do, just to update to new function signatures


oleavr:
Thanks! :raised_hands:

PR #269, opened 2022-10-17
Add GitHub Actions

yotamN:
I should probably organize it a bit better, with test matrix and such but it works :)


yotamN:
The arm tests are blocked by https://github.com/ReactiveCircus/android-emulator-runner/issues/280 so I think we should merge it as is for now.

PR #270, opened 2022-10-29
Use symbols for unique property names 

yotamN:
This is only a proposal of how to do it. Using only `Symbol` would be a bit verbose, so maybe only exposing it as a fallback would be okay?

I also used the global registry for symbols, we can also use a unique global `Symbol` variable if you think it would be better.

### DIFF
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -186,39 +186,94 @@ class ClassFactory {
   _make (name, getClassHandle, env) {
     const C = makeClassWrapperConstructor();
     const proto = Object.create(Wrapper.prototype, {
-      $n: {
+      [Symbol.for('n')]: {
         value: name
       },
-      $C: {
+      $n: {
+        get () {
+          return this[Symbol.for('n')];
+        }
+      },
+      [Symbol.for('C')]: {
         value: C
       },
-      $w: {
+      $C: {
+        get () {
+          return this[Symbol.for('C')];
+        }
+      },
+      [Symbol.for('w')]: {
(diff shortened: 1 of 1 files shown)


yotamN:
I've added symbols in the class-factory file but we are not using them yet inside the code itself.  
I'm waiting for feedback to see that it's okay to proceed.

lib/class-factory.js
```diff
@@ -15,6 +15,7 @@ const {
   getArrayType,
   makeJniObjectTypeName
 } = require('./types');
+const memoize = require('./memoize');
```
oleavr, line 18: Imports should be sorted alphabetically by module name.

lib/class-factory.js
```diff
@@ -186,39 +187,94 @@ class ClassFactory {
   _make (name, getClassHandle, env) {
     const C = makeClassWrapperConstructor();
     const proto = Object.create(Wrapper.prototype, {
-      $n: {
+      [Symbol.for("n")]: {
```
oleavr, line 190: Strings should be single-quoted.

oleavr (changes requested):
Thanks! Only some style changes needed:

lib/class-factory.js
yotamN: Fixed

lib/class-factory.js
yotamN: Fixed, should not have been there

oleavr:
Thanks! 🙌

PR #272, opened 2022-10-29
class-factory: Fix toString for overloaded methods

### LINKED ISSUE #271: Classes that override toString with additional arguments break the toString of Frida
For example, the class `android.icu.text.UTF16` override the `toString` method privately with an implementation that requires an integer: https://android.googlesource.com/platform/external/icu/+/2ae1300/android_icu4j/src/main/java/android/icu/text/UTF16.java#2708

The parent method is not find because when we call toString it only iterates on the current class. Of course, we could access the parent but we already have a patch for toString on non instance classes to make it easier: https://github.com/frida/frida-java-bridge/blob/main/lib/class-factory.js#L1402

I think we should do that same for this use case

yotamN:
Fixes #271 
Depends on #275

### DIFF
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -1409,6 +1409,10 @@ dispatcherPrototype = Object.create(Function.prototype, {
         return method.apply(receiver, args);
       }
 
+      if (this.methodName === 'toString') {
+        return `<class: ${receiver.$n}>`;
+      }
+
       throwOverloadError(this.methodName, this.overloads, 'argument types do not match any of:');
     }
   }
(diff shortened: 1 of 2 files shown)


lib/class-factory.js
```diff
@@ -1409,6 +1409,10 @@ dispatcherPrototype = Object.create(Function.prototype, {
         return method.apply(receiver, args);
       }
 
+      if (this.methodName === 'toString') {
+          return `<class: ${receiver.$n}>`;
```
oleavr, line 1413: Indentation should be two spaces.

oleavr (changes requested):
Thanks! Only one style issue and this should be good to merge:

lib/class-factory.js
yotamN: Fixed

oleavr:
Thanks! 🙌

PR #275, opened 2022-12-23
Fix CI

yotamN:
There were several problems with the current CI, hopefully now everything will work in a predictable way


oleavr:
Thanks!

PR #278, opened 2023-01-27
class-factory: Add toString method to types and field

yotamN:
In general there is too much [object object]

### DIFF
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -1986,7 +1986,7 @@ Object.defineProperties(Field.prototype, {
 \tfieldReturnType: ${this.fieldReturnType},
 \tvalue: ${this.value},
 }`;
-      return multilineString.split('\n').map(l => l.length > 200 ? l.slice(0, l.indexOf(" ")+1) + "...," : l).join('\n');
+      return multilineString.split('\n').map(l => l.length > 200 ? l.slice(0, l.indexOf(' ') + 1) + '...,' : l).join('\n');
     },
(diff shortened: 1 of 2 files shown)


lib/class-factory.js
```diff
@@ -1972,6 +1972,22 @@ Object.defineProperties(Field.prototype, {
     get () {
       return this._p[2];
     }
+  },
+  toString: {
+    enumerable: true,
+    value() {
```
oleavr, line 1978:
```suggestion
    value () {
```

lib/class-factory.js
```diff
@@ -1972,6 +1972,22 @@ Object.defineProperties(Field.prototype, {
...
+      if (inlineString.length < 200) {
+        return inlineString;
+      }
+      const multilineString = `Java.Field{
+\tholder: ${this.holder},
+\tfieldType: ${this.fieldType},
+\tfieldReturnType: ${this.fieldReturnType},
+\tvalue: ${this.value},
+}`;
+      return multilineString.split('\n').map(l => l.length > 200 ? l.slice(0, l.indexOf(" ")+1) + "...," : l).join('\n');
```
oleavr, line 1989:
```suggestion
      return multilineString.split('\n').map(l => l.length > 200 ? l.slice(0, l.indexOf(' ') + 1) + '...,' : l).join('\n');
```

oleavr (changes requested):
Thanks! 🙌

PR #281, opened 2023-02-05
*bug fix* wrong overload invocation of overloaded functions - new idea

Roee-BY:
As a continuation of #247 , what if instead of throwing an error we will just warn the user and notify him which overload was selected.

the original pull request message:
"the function invocation mechanism was choosing the first compatible func according the the args and available overloads but there was a bug since some types that are available in java are unified into one type in JS for example JS number can be int double float or long hence choosing the first might cause a problem (in my case an android app crashes due to the incorrect overload). instead of the current mechanism i have implemented a mechanism that goes over all of the overloads and if there is more then one matching overload it invokes the method throwIfDispatcherAmbiguous(dispatcher) which will alert the user about the multiple a available overloads and ask his to choose which one to use."

### DIFF
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -1390,7 +1390,9 @@ dispatcherPrototype = Object.create(Function.prototype, {
       if (methodsThatCanBeInvoked.length === 0) {
         throwOverloadError(this.methodName, this.overloads, 'argument types do not match any of:');
       } else if (methodsThatCanBeInvoked.length > 1) {
-        throwIfDispatcherAmbiguous(this);
+        console.warn("Multiple overloads fit the invocation of " + receiver + 
+        ",\n The overload that will be used is: " + methodsThatCanBeInvoked[0] +
+        ",\n To avoid this warining please choose one of the follwoing overloads and invoke it specifically: " + methodsThatCanBeInvoked);
       }
       const method = methodsThatCanBeInvoked[0];
 
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -1386,40 +1386,38 @@ dispatcherPrototype = Object.create(Function.prototype, {
   invoke: {
     value (receiver, args) {
       const overloads = this._o;
-
-      const isInstance = receiver.$h !== null;
-      
-      const MethodsThatCanBeInvoked=[];
+      const isInstance = receiver.$h !== null; 
+      const methodsThatCanBeInvoked=[];
 
       for (let i = 0; i !== overloads.length; i++) {
         const method = overloads[i];
 
         if (method.canInvokeWith(args)) {
-          if (MethodsThatCanBeInvoked.length > 0) {
+          if (methodsThatCanBeInvoked.length > 0) {
             throwIfDispatcherAmbiguous(this);
           }
           
-          MethodsThatCanBeInvoked.push(method);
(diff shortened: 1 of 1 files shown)


lib/class-factory.js
```diff
@@ -1386,30 +1386,28 @@ dispatcherPrototype = Object.create(Function.prototype, {
   invoke: {
     value (receiver, args) {
       const overloads = this._o;
+      const methodsThatCanBeInvoked = overloads.filter(method => method.canInvokeWith(args));
+      if (methodsThatCanBeInvoked.length === 0) {
+        throwOverloadError(this.methodName, this.overloads, 'argument types do not match any of:');
```
oleavr, line 1391: Should we use `this.overloads` here when we have the `overloads` local variable right above us? (If it's the same -- I can't remember, but I think so.)

lib/class-factory.js
```diff
@@ -1386,30 +1386,28 @@ dispatcherPrototype = Object.create(Function.prototype, {
...
+      } else if (methodsThatCanBeInvoked.length > 1) {
+        console.warn("Multiple overloads fit the invocation of " + receiver +
```
oleavr, line 1393: There's no need to concatenate strings, `console.warn()` will add a space between the stringified value of each argument.

lib/class-factory.js
```diff
@@ -1386,30 +1386,28 @@ dispatcherPrototype = Object.create(Function.prototype, {
   invoke: {
     value (receiver, args) {
       const overloads = this._o;
+      const methodsThatCanBeInvoked = overloads.filter(method => method.canInvokeWith(args));
+      if (methodsThatCanBeInvoked.length === 0) {
+        throwOverloadError(this.methodName, this.overloads, 'argument types do not match any of:');
+      } else if (methodsThatCanBeInvoked.length > 1) {
+        console.warn("Multiple overloads fit the invocation of " + receiver + 
+        ",\n The overload that will be used is: " + methodsThatCanBeInvoked[0] +
+        ",\n To avoid this warining please choose one of the follwoing overloads and invoke it specifically: " + methodsThatCanBeInvoked);
```
oleavr, line 1395:
```suggestion
        ",\n To avoid this warning please choose one of the following overloads and invoke it specifically: " + methodsThatCanBeInvoked);
```

lib/class-factory.js
```diff
@@ -1386,30 +1386,28 @@ dispatcherPrototype = Object.create(Function.prototype, {
...
-            return `<class: ${receiver.$n}>`;
-          }
-
-          throw new Error(name + ': cannot call instance method without an instance');
+        if (name === 'toString') {
+          return `<class: ${receiver.$n}>`;
         }
 
-        return method.apply(receiver, args);
+        throw new Error(name + ': cannot call instance method without an instance');
```
oleavr, line 1407: Would be good to use a template string here.

oleavr (changes requested):
Thanks! Apologies for taking so long to get to this, I've been spread too thin.

I think this is good approach, but this needs test coverage before it can be merged. We need to ensure that we cover the different cases.

lib/class-factory.js
Roee-BY: you are right, `this.overloads` and the `overloads ` local variable have the same value.

PR #283, opened 2023-03-18
Add toString method for overloads

oriori1703:
Like the title says 😄

### DIFF
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -1647,6 +1647,12 @@ methodPrototype = Object.create(Function.prototype, {
         env.popLocalFrame(NULL);
       }
     }
+  },
+  toString: {
+    enumerable: true,
+    value (){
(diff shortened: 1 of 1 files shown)


lib/class-factory.js
```diff
@@ -1647,6 +1647,12 @@ methodPrototype = Object.create(Function.prototype, {
         env.popLocalFrame(NULL);
       }
     }
+  },
+  toString: {
+    enumerable: true,
+    value (){
+        return `function ${this.methodName}(${this.argumentTypes.map(t => t.className).join(', ')}): ${this.returnType.className}`
```
oleavr, line 1654:
```suggestion
    value () {
      return `function ${this.methodName}(${this.argumentTypes.map(t => t.className).join(', ')}): ${this.returnType.className}`;
```

oleavr (changes requested):
Thanks! 🙌

Issue #288, opened 2023-05-18
[Question] How do use a local version of this library in a local Frida build?


gsingh93:
I've fixed a bug in `frida-java-bridge` that gets it working on the new Android 14 that I would like to upstream to this repo. I had a very hard time testing this fix though, as I needed to patch [bindings/gumjs/generate-runtime.py](https://github.com/frida/frida-gum/blob/ac1a1574d80349f8d19d41f86a5c8ef0e4f04575/bindings/gumjs/generate-runtime.py#L44-L51) in frida-gum to use my local copy of this repo.

Is there any easier way to do this?

PR #289, opened 2023-05-18
Make _ZNK3art6Thread13DecodeJObjectEP8_jobject optional

gsingh93:
This will fix https://github.com/frida/frida/issues/2468

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -353,7 +353,8 @@ function _getApi () {
           '_ZN3art3Dbg9gRegistryE',
           '_ZN3art3jni12JniIdManager14DecodeMethodIdEP10_jmethodID',
           '_ZN3art11interpreter18GetNterpEntryPointEv',
-          '_ZN3art7Monitor17TranslateLocationEPNS_9ArtMethodEjPPKcPi'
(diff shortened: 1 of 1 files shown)


gsingh93:
As this is only used for Android 6 and lower, I think making it optional makes sense instead of having a fallback.

And as a sidenote, I think the current system of looking for these methods is prone to break often. I would suggest instead of attempting to find every method not marked as optional, we just find the version of android we're connecting to, and then have a list of APIs that were added/removed from each version. Then, if we have an API in the "removed" list for Android 7, it will still be required for Android 6 and lower, but not required for anything above Android 6.

gsingh93:
@oleavr can you run the checks that require approval?

AAGaming00:
Might be slightly off-topic, but how would I build a frida-server binary that uses this?

gsingh93:
The build instructions are the normal ones for Frida, except to use the local version of frida-java-bridge I had to make this modification to frida-gum: https://github.com/gsingh93/frida-gum/commit/dabfe43ecbeb1a83a02b4628e8f85c723accc206

I've asked in #288 what the right way to use a local version of a library is.

PR #307, opened 2024-01-16
i#306: Fix hooking art_quick_resolution_trampoline

cr4zyserb:
Fixes issue when using --enable-optimizations on newer ARTs where quick entrypoints can be art_quick_resolution_trampoline


oleavr:
Thanks! :metal:

PR #311, opened 2024-04-03
Add support for instantiating classes that do not have a constructor

AeonLucid:
With this PR you can instantiate java classes that do not have a constructor, by using the `java/lang/Object->init()` method.
I'm not sure if we need to handle more edge cases like the `isInterface` check.


johansci:
@oleavr can you merge this PR?

oleavr:
Yay, this is great. Thanks! 🙌

PR #318, opened 2024-05-21
fix(android): set DexFile to read-only in Java.registerClass()

pandasauce:
As of Android 14, apps with targetSdk >= 34 are not allowed to have writable permissions on dynamically loaded Dex files.

https://developer.android.com/about/versions/14/behavior-changes-14#safer-dynamic-code-loading

Attempting to call `Java.registerClass()` with such apps results in an exception being thrown:

```
java.lang.SecurityException: Writable dex file '/data/data/my.ebin.app/frida4706282976511766825.dex' is not allowed.
```

This removes writable permissions from the temporary DexFile created by `Java.registerClass()` after writing to it and before it gets loaded.

### DIFF
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -2193,6 +2193,7 @@ class DexFile {
     const file = new File(filePath, 'w');
     file.write(buffer.buffer);
     file.close();
+    setReadOnlyDex(filePath, factory);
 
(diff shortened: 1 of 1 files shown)


oleavr:
Nice!! Thanks! 🙌

Issue #296, opened 2023-07-19
ClassNotFoundException when Java.registerClass() declaring more than 2 fields


LanceaKing:
This code works normally:
```javascript
Java.perform(() => {
    const MyRunnable = Java.registerClass({
        name: "re.frida.MyRunnable",
        implements: [Java.use("java.lang.Runnable")],
        fields: {
            a: "int",
            b: "int",
            // c: "int"
        },
        methods: {
            run() {
                // ...
            }
        }
    })
})
```

This code raise `java.lang.ClassNotFoundException`:
```javascript
Java.perform(() => {
    const MyRunnable = Java.registerClass({
        name: "re.frida.MyRunnable",
        implements: [Java.use("java.lang.Runnable")],
        fields: {
            a: "int",
            b: "int",
            c: "int"
        },
        methods: {
            run() {
                // ...
            }
        }
    })
})
```

```
Error: java.lang.ClassNotFoundException: Didn't find class "re.frida.MyRunnable" on path: DexPathList[[],nativeLibraryDirectories=[/system/lib64, /system_ext/lib64]]
```

LanceaKing:
@oleavr Frida version is 16.1.2

eybisi:
I encountered the same issue. Introducing 3rd field, throws ClassNotFoundException

PR #322, opened 2024-06-05
fix: sort field items according to dex_file_verifier.cc

eybisi:
registerClass creates wrong dex file when there are more than 2 fields #296 

This commit fixes two errors:
- field_ids items are not sorted. 
- field_idx_diff wrongly calculated. (0,1,2,3) needs to be (0,1,1,1)

Since we create 1 dex file for each registerClass, we don't need to keep track of field_idx_diff. (okay to begin from 0)

### DIFF
--- a/lib/mkdex.js
+++ b/lib/mkdex.js
@@ -655,7 +655,7 @@ function computeModel (classes) {
       stringToIndex[fieldName]
     ];
   });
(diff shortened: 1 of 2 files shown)


oleavr:
Thanks! :raised_hands:

Issue #323, opened 2024-06-05
SIGSEGV in artQuickGenericJniTrampoline while hooking java methods


matbrik:
Hooking a java method in system_server on a Samsung Android 13 S23 plus and A33 with the last updates causes a SIGSEGV/SEGV_MAPERR crash. On a S21 5G I cannot reproduce it.
My hypothesis is a change in libart.so but looking at the source code and diffing the binaries I couldn't find a reason.
p.s. not all methods seems to trigger the crash
sha1sum: 7c9ef90838717ac4d792139f8b1f7ca9692d018e  /apex/com.android.art@`341711000`/lib64/libart.so
frida-gadget 16.2.5

crashlog
```

06-05 11:55:59.313  2457  2457 I frida   : Debug1
06-05 11:55:59.318  2457  2457 I frida   : ComputerEngine: <class: com.android.server.pm.ComputerEngine>
...
...
...
06-05 11:56:07.354  2457  3623 F libc    : Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x1b96c0 in tid 3623 (binder:2457_4), pid 2457 (system_server)
...
...
...
06-05 11:56:08.544 18081 18081 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
06-05 11:56:08.544 18081 18081 F DEBUG   : Build fingerprint: 'samsung/dm2qxeea/dm2q:13/TP1A.220624.014/S916BXXS3AWIF:user/release-keys'
06-05 11:56:08.544 18081 18081 F DEBUG   : Revision: '13'
06-05 11:56:08.544 18081 18081 F DEBUG   : ABI: 'arm64'
06-05 11:56:08.544 18081 18081 F DEBUG   : Processor: '5'
06-05 11:56:08.544 18081 18081 F DEBUG   : Timestamp: 2024-06-05 11:56:07.512826284+0200
06-05 11:56:08.544 18081 18081 F DEBUG   : Process uptime: 1136s
06-05 11:56:08.544 18081 18081 F DEBUG   : Cmdline: system_server
06-05 11:56:08.544 18081 18081 F DEBUG   : pid: 2457, tid: 3623, name: binder:2457_4  >>> system_server <<<
06-05 11:56:08.544 18081 18081 F DEBUG   : uid: 1000
[...]

PR #326, opened 2024-07-16
android: Handle correctly the CMC GC strategy

matbrik:
Fixes https://github.com/frida/frida-java-bridge/issues/323.

The fix in https://github.com/frida/frida-java-bridge/pull/325 had 2 problems:
- with the Google Play System Updates the base version of libart can differ from the Android version( i.e. Android 12 device with libart 34)
- the function we were hooking in the MarkCompact GC was not stable enough causing crashes after some time. [tombstone_01.txt](https://github.com/user-attachments/files/16249961/tombstone_01.txt)

The proposed solution to the first point is to check if an exported symbol introduced in libart 34 is present.

The solution for the second problem is to hook the onEnter of art::Thread::RunFlipFunction
- this function is called also during in the ConcurrentCopying GC and in my tests didn't cause troubles but to avoid compatibility problems with untested version I've reintroduced the check on the GC strategy;
- the downside of hooking art::Thread::RunFlipFunction is that it is called several times (tens) every run of the GC but anyother place I've tried was causing crashes.

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -618,6 +618,7 @@ function _getArtRuntimeSpec (api) {
 
   const apiLevel = getAndroidApiLevel();
   const codename = getAndroidCodename();
+  const isApiLevel34OrApexEquivalent = Module.findExportByName('libart.so', '_ZN3art7AppInfo29GetPrimaryApkReferenceProfileEv') !== null;
 
   let spec = null;
 
@@ -645,7 +646,7 @@ function _getArtRuntimeSpec (api) {
         const threadListOffset = internTableOffset - pointerSize;
 
         let heapOffset;
-        if (apiLevel >= 34) {
+        if (isApiLevel34OrApexEquivalent) {
           heapOffset = threadListOffset - (9 * pointerSize);
         } else if (apiLevel >= 24) {
           heapOffset = threadListOffset - (8 * pointerSize);
--- a/lib/android.js
+++ b/lib/android.js
@@ -1808,6 +1808,9 @@ on_leave_gc_concurrent_copying_copying_phase (GumInvocationContext * ic)
       Gc: {
         copyingPhase: {
           onLeave: cm.on_leave_gc_concurrent_copying_copying_phase
+        },
+        runFlip: {
+          onEnter: cm.on_leave_gc_concurrent_copying_copying_phase
         }
       }
     }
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks! Just applied some style tweaks and refactored a little bit. It would take me some time to be able to test this code though, so was wondering if you could take it for a quick spin? :)

matbrik:
I've tested on 3 devices and it looks like it is working fine

oleavr:
Thanks! 🙌

esauvisky:
This fails in Android 15/SDK35 though:

```
Error: libart.so: unable to find export '_ZNK3art2gc4Heap15MayUseCollectorENS0_13CollectorTypeE'
    at Function.value [as getExportByName] (frida/runtime/core.js:250:1)
    at Ft (frida/node_modules/frida-java-bridge/lib/android.js:837:1)
    at Tt (frida/node_modules/frida-java-bridge/lib/android.js:762:1)
    at mn.replace (frida/node_modules/frida-java-bridge/lib/android.js:1331:1)
    at Function.set [as implementation] (frida/node_modules/frida-java-bridge/lib/class-factory.js:1189:1)
    at Function.set [as implementation] (frida/node_modules/frida-java-bridge/lib/class-factory.js:1103:1)
    at agent/utils.js:65:51
    at c.perform (frida/node_modules/frida-java-bridge/lib/vm.js:12:1)
    at _.perform (frida/node_modules/frida-java-bridge/index.js:205:1)
```

This is because `MayUseCollector` is no longer exported by libart. Furthermore, the previous strategy is still ineffective since none of the three methods below are exported either:
- `_ZN3art2gc9collector17ConcurrentCopying12CopyingPhaseEv`
- `_ZN3art2gc9collector17ConcurrentCopying12MarkingPhaseEv`
- `_ZN3art2gc9collector11MarkCompact15CompactionPhaseEv`

The function `RunFlipFunction` **is** exported, but apparently they removed a parameter, so its signature has changed from `_ZN3art6Thread15RunFlipFunctionEPS0_b` to `_ZN3art6Thread15RunFlipFunctionEPS0_`.

pig837:
> Fixes #323.
> 
> The fix in #325 had 2 problems:
> 
> * with the Google Play System Updates the base version of libart can differ from the Android version( i.e. Android 12 device with libart 34)
> * the function we were hooking in the MarkCompact GC was not stable enough causing crashes after some time. [tombstone_01.txt](https://github.com/user-attachments/files/16249961/tombstone_01.txt)
> 
> The proposed solution to the first point is to check if an exported symbol introduced in libart 34 is present.
> 
> The solution for the second problem is to hook the onEnter of art::Thread::RunFlipFunction
> 
> * this function is called also during in the ConcurrentCopying GC and in my tests didn't cause troubles but to avoid compatibility problems with untested version I've reintroduced the check on the GC strategy;
> * the downside of hooking art::Thread::RunFlipFunction is that it is called several times (tens) every run of the GC but anyother place I've tried was causing crashes.

[Related Issues]
- https://github.com/frida/frida/issues/2899
- https://github.com/frida/frida-java-bridge/issues/323

Thank you, you saved me!

Hooking art::Thread::RunFlipFunction function is work well for me.

But, Frida 16.4.7 don't have the feature bypassing RunFlipFunction.
Hopefully a new version will be released that addresses this issue.

NgoHuy:
the fix is incomplete and introduce new bug, I added some comments, I also create new pull to correct it (I didnot add #330  fix, it should be merged seperately)

PR #328, opened 2024-07-23
Feature: Will bridge Windows Java (OpenJDK17)

FrankSpierings:
This change will make the `jvm.js` code be able to bridge Java OpenJDK 17 on Windows. This was tested on Windows 10, using `jdk-17.0.11+9` from https://learn.microsoft.com/en-gb/java/openjdk/download and the separately downloaded and extracted debug symbols.

The code is not pretty, but I've tried to keep as close to the original as possible.

Note the bug that I discovered while performing the tests; https://github.com/frida/frida-gum/issues/811. This means that the target process needs to be restarted after detaching Frida, for the Java bridge (resolving of Symbols) to work.

This code currently does not work with OpenJDK21!

This code requires the 'manual' placement of the debug symbols, otherwise `jvm.ts` will be unable to find the functions it requires.

### DIFF
--- a/lib/jvm.js
+++ b/lib/jvm.js
@@ -51,184 +51,226 @@ function _getApi () {
     flavor: 'jvm'
   };
 
-  const pending = [{
-    module: vmModule.path,
-    functions: {
-      JNI_GetCreatedJavaVMs: ['JNI_GetCreatedJavaVMs', 'int', ['pointer', 'int', 'pointer']],
+  const pending = Process.platform === 'windows'
+    ? [{
+        module: vmModule.path,
+        functions: {
+          JNI_GetCreatedJavaVMs: ['JNI_GetCreatedJavaVMs', 'int', ['pointer', 'int', 'pointer']],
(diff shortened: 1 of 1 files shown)


CDuPlooy:
This is super cool, I managed to get it somewhat working by following your instructions and this [gist](https://gist.github.com/oleavr/cae76c895eb7d227216ed3ffe9dbbeb3). The example frida script from the gist did work, however if I try to write a hook for a basic class, I get the following:

```
Unable to make thread_from_jni_environment() helper for the current architecture
```

This was on an x64 machine, and is as far as I got with debugging; Curious if it works on your side, also happy to help out where possible :)

FrankSpierings:
You are right. I have basic functionality working, so you can create a new instance of a class and execute its methods, but I also noticed that once you attempt to start interception it breaks. Sorry, I should have tested more thoroughly before I went on vacation. I'll attempt to fix this functionality, although it appears more daunting then initially thought.

oleavr:
Thanks! 🙌

PR #330, opened 2024-07-30
android: fix bug where MayUseCollector is not available on some custom rom

thinhbuzz:
fix `Error: libart.so: unable to find export '_ZNK3art2gc4Heap15MayUseCollectorENS0_13CollectorTypeE`
eg: https://github.com/frida/frida/issues/2925
original change: https://github.com/frida/frida-java-bridge/pull/326

PR #332, opened 2024-08-15
The fix for incomplete #326

NgoHuy:
When the fix is merged, it introduces new issue, exportName is never null, it always attaches to null functions, then crash our server.


NgoHuy:
the fix is merged at #330 then we can close this PR if #330 was merged

PR #340, opened 2024-10-28
Fix JNI ERROR

123edi10:
After executing `Java.enumerateLoadedClassesSync()` on certain devices and within multiple apps I got this error:

> signal 6 (SIGABRT) error, with code -6 (SI_TKILL) and fault address --------. 
The abort message reads: "JNI ERROR (app bug): global reference table overflow (max=51200). 
Global reference table dump: Last 10 entries (of 51200):".

This solution tested in several devices and work perfect.

### DIFF
--- a/index.js
+++ b/index.js
@@ -222,7 +222,6 @@ class Runtime {
         try {
           const className = env.getClassName(handle);
           callbacks.onMatch(className, handle);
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks! 🙌

Issue #309, opened 2024-02-08
bug in function instrumentArtMethodInvocationFromInterpreter


fvrmatteo:
When running Android 13 (API 33), the function [instrumentArtMethodInvocationFromInterpreter](https://github.com/frida/frida-java-bridge/blob/main/lib/android.js#L1841) function sets the `artInterpreterDoCallExportRegex` variable to the second regex (`/^_ZN3art11interpreter6DoCallILb[0-1]ELb[0-1]EEEbPNS_9ArtMethodEPNS_6ThreadERNS_11ShadowFrameEPKNS_11InstructionEtPNS_6JValueE$/`), although on my Pixel 6 running Android 13 I can see that the `DoCall` symbols present on `libart.so` would actually match the third regex (`/^_ZN3art11interpreter6DoCallILb[0-1]EEEbPNS_9ArtMethodEPNS_6ThreadERNS_11ShadowFrameEPKNS_11InstructionEtbPNS_6JValueE$/`).

It might be that originally the `apiLevel <= 33` condition wasn't a bug, but since `libart.so` gets updated via APEX, the symbols might have changed in the meantime, making the condition invalid.

If the bug is confirmed, an idea could be to match against all the regexes and keep only the discovered symbols, but I cannot comment on the performance impact.

lisbonresearch:
+1 on this issue. I have a device with Android 13 and com.android.art version 341311100. This means that the code of `instrumentArtMethodInvocationFromInterpreter` goes to the second branch (because the API level is 33), but the correct symbol cannot be found, because the signature of `DoCall` for art version 34nnnnnnn is:

```
template<bool is_range>
bool DoCall(ArtMethod* called_method,
            Thread* self,
            ShadowFrame& shadow_frame,
            const Instruction* inst,
            uint16_t inst_data,
            bool string_init,
            JValue* result);
```

which corresponds to the mangled name `_ZN3art11interpreter6DoCallILb[0-1]EEEbPNS_9ArtMethodEPNS_6ThreadERNS_11ShadowFrameEPKNS_11InstructionEtbPNS_6JValueE`.

If there's no easier way of finding the version of art (one could, for example, process the first bytes of the protobuf-encoded file `apex_manifest.pb` but I'm not sure if this is an interesting approach), maybe the solution is to test all the regexes as mentioned previously. In any case, the current code assumes that the version of art is the same as the API level of Android, which isn't always the case.

mrexodia:
[...]

PR #343, opened 2025-01-28
android: Support new ART on compatible Android versions

kaftejiman:
Recent versions of ART can run on older Android versions. Because some of the ClassLinker’s specifications and method-invocation logic are tied exclusively to the API level, this pull request also considers the ART level to ensure compatibility

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -151,6 +151,11 @@ function _getApi () {
     addLocalReference: null
   };
 
+  temporaryApi.isApiLevel34OrApexEquivalent = isArt && (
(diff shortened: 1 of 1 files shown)


kaftejiman:
Fixes frida/frida-java-bridge#309

oleavr:
Thank you! :raised_hands:

lib/android.js
```diff
@@ -944,6 +950,8 @@ function tryGetArtClassLinkerSpec (runtime, runtimeSpec) {
 
   if (spec !== null) {
     cachedArtClassLinkerSpec = spec;
+  } else {
+    throw new Error('Unable to determine ClassLinker field offsets');
```
thinhbuzz, line 954: @oleavr Can we revert this change? it causes problem #347

PR #344, opened 2025-02-22
Update the Frida version for the test to 16.6.6

oriori1703:
This was needed after 33307d4 beacuse it uses the new `Module.findSymbolbyName` that was introduced in v16.6.0


oriori1703:
This seems to fix the main issue with the test, but now the tests are segfaulting on x86.

I will try to debug it if I have time this week.
until then, @oleavr do you have any ideas?

PR #349, opened 2025-04-25
fix handling conditional branch when recompiling for arm32

Rwkeith:
bne.w instruction is now handled in arm recompile routine resulting in successful art thread state transition for cases where this instruction appears in the code being recompiled.


oleavr:
Thanks! 🙌🏻

Issue #333, opened 2024-08-16
Hooking functions on JDK17 (HotSpot) crashes the instrumented application


c8rri3r:
## Summary
Hooking a function on JDK17 (HotSpot) causes the instrumented application to crash. Other features such as searching methods with `Java.enumerateMethods(query)` work as expected as far as I can tell. This happens each and every time a Java function is hooked. Given this sample application:

```java
public class App 
{
    public static void main( String[] args )
    {
        System.out.println("[*] Checking status of someTest()...");
        try {
            while(true) {
                boolean result = someTest();
                if(result) {
                    System.out.println("[*] Congrats, result has been changed!");
                }
                TimeUnit.SECONDS.sleep(10);
            }
        } catch (InterruptedException e) { }
    }
    public static boolean someTest() {
        return false;
    }
}
```

Hooking the `someTest()` function with this  instrumentation code will cause the crash:
```javascript
Java.perform(function () {
    console.log("Is Java Available: " + Java.available)
    var MyClass = Java.use('com.frida.App');

    // Hook the "someTest" function
    MyClass.someTest.implementation = function () {
        console.log('someTest() function called');
        return true;
    };
});
```
The error message (trimmed for brevity) doesn't immediately indicate where the issue is.
```bash
└─$ $JAVA_HOME/bin/java -jar target/frida-java-demo-1.0-SNAPSHOT-jar-with-dependencies.jar
[*] Checking status of someTest()...
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x0000000000000000, pid=4713, tid=4714
#
# JRE version: OpenJDK Runtime Environment Temurin-17.0.12+7 (17.0.12+7) (build 17.0.12+7)
# Java VM: OpenJDK 64-Bit Server VM Temurin-17.0.12+7 (17.0.12+7, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# j  com.frida.App.main([Ljava/lang/String;)V+8
...
```

The Frida console provides a more descriptive error message:

```javascript
Is Java Available: true
Error: Unable to make thread_from_jni_environment() helper for the current architecture
    at <anonymous> (frida/node_modules/frida-java-bridge/lib/jvm.js:232)
    at <anonymous> (frida/node_modules/frida-java-bridge/lib/jvm.js:276)
[...]

PR #350, opened 2025-04-26
Fixing issue #333 with JDK17 method replacement

c8rri3r:
Patch to fix the issue with replacing JDK17 methods. See https://github.com/frida/frida-java-bridge/issues/333 for the full details.


oleavr:
Thanks! 🙌

PR #351, opened 2025-05-17
Support changes to art Class layout in android 16

tbodt:
https://android-review.googlesource.com/c/platform/art/+/3380108

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -965,27 +965,36 @@ export function getArtClassSpec (vm) {
     return null;
   }
 
-  if (apiLevel < 24) {
+  if (apiLevel >= 34) {
+    return {
+      offset: {
+        ifields: 0x28,
+        methods: 0x28 + 0x8,
+        sfields: 0,
+        copiedMethodsOffset: 0x6c,
+      }
+    };
+  } else if (apiLevel >= 26) {
+    return {
+      offset: {
+        ifields: 0x28,
+        methods: 0x28 + 0x8,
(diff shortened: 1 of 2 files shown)


tbodt:
I have not actually been able to test this because I can't get frida-compile to read es modules - i developed this against an older release without esm and ported it to head this PR, so i've tested it but not in this exact verison.

oleavr:
Thanks! 🙌

lib/android.js
```diff
@@ -965,27 +965,36 @@ export function getArtClassSpec (vm) {
...
 
-  if (apiLevel < 24) {
-    return null;
-  }
-
-  let base, cmo;
-  if (apiLevel >= 26) {
-    base = 40;
-    cmo = 116;
+  if (apiLevel >= 34) {
```
thinhbuzz, line 968: this will cause error `TypeError: t.use(...).currentApplication is not a function` on android 14 and 15, this condition needs to be `>= 36` - only applies to Android 16 and above

@tbodt Can you make a PR for it?

cc: @oleavr 

```
TypeError: t.use(...).currentApplication is not a function
    at frida/node_modules/frida-java-bridge/index.js:214:1
    at c.perform (frida/node_modules/frida-java-bridge/lib/vm.js:12:1)
    at _.performNow (frida/node_modules/frida-java-bridge/index.js:211:1)
    at _.scheduleOnMainThread (frida/node_modules/frida-java-bridge/index.js:175:1)
    
    ```

PR #356, opened 2025-06-12
Fix wrapper alloc

### LINKED ISSUE #353: class-factory.js $alloc Symbol.for('alloc')
```js
  [Symbol.for('alloc')]: {
    enumerable: false,
    value () {
      const env = vm.getEnv();
      const h = this.$borrowClassHandle(env);
      try {
        const obj = env.allocObject(h.value);
        const factory = this.$f;
        return factory.cast(obj, this);
      } finally {
        h.unref(env);
      }
    }
  },
  $alloc: {
    enumerable: true,
    get () {
      return this[Symbol.for('$alloc')];
    }
  },
```

When call obj.$alloc, throw `TypeError: not a function`

Sorry For no PR.

yotamN:
While migrating to using symbols, I migrated alloc with a wrong name.

Fixes #353

### DIFF
--- a/lib/class-factory.js
+++ b/lib/class-factory.js
@@ -899,7 +899,7 @@ Object.defineProperties(Wrapper.prototype, {
   $alloc: {
     enumerable: true,
     get () {
-      return this[Symbol.for('$alloc')];
+      return this[Symbol.for('alloc')];
     }
   },
   [Symbol.for('init')]: {


oriori1703 (approved):
LGTM

oleavr:
Thanks! :raised_hands:

PR #359, opened 2025-07-11
Fix system_server crashes - automatically detect offsets for ART class spec

AeonLucid:
This PR requires a new Frida build because it affects the internal system_server agent.

Closes
- https://github.com/frida/frida/issues/3533
  - https://github.com/frida/frida/issues/3525
  - https://github.com/frida/frida/issues/3527
  - https://github.com/frida/frida-core/issues/1165
  - https://github.com/ViRb3/magisk-frida/issues/62
 
---

As you can see this is a widespread issue. 

The issue is that the current code only accounts for the ART class layout based on the android sdk.
However, `libart.so` is shipped with the google security updates, that way you can have an Android 16 layout on Android 15.
This is why `pm uninstall com.google.android.art` is a "fix", although temporarily.

I fixes this by automatically detecting the offsets for the art class spec. By using a known class `java/lang/Integer` we can search for known fields / methods and find the necessary offsets that way.

PR #361, opened 2025-07-20
Use stable fields for ART class spec offset detection

AeonLucid:
Addresses crashes commented under the previous pull request https://github.com/frida/frida-java-bridge/pull/359.
These fields shouldn't be optimised away.

Edit: Also added an extra check for the art array length. Sometimes it would read a negative number when trying offsets, and it would infinitely loop.

### DIFF
--- a/lib/android.js
+++ b/lib/android.js
@@ -985,7 +985,7 @@ export function getArtClassSpec (vm) {
       }
 
       const length = (lengthSize === 4) ? header.readU32() : header.readU64().valueOf();
-      if (length === 0) {
+      if (length <= 0) {
         return null;
       }
 
@@ -1002,7 +1002,8 @@ export function getArtClassSpec (vm) {
           return false;
         }
 
-        for (let i = 0; i !== Math.max(0, Math.min(artArray.length, info.artArrayMax)); i++) {
+        const artArrayEnd = Math.min(artArray.length, info.artArrayMax);
+        for (let i = 0; i !== artArrayEnd; i++) {
           const fieldPtr = artArray.data.add(i * info.artArrayEntrySize);
           if (fieldPtr.equals(needle)) {
             return true;
--- a/lib/android.js
+++ b/lib/android.js
@@ -1002,7 +1002,7 @@ export function getArtClassSpec (vm) {
           return false;
         }
 
-        for (let i = 0; i !== Math.min(artArray.length, info.artArrayMax); i++) {
+        for (let i = 0; i !== Math.max(0, Math.min(artArray.length, info.artArrayMax)); i++) {
           const fieldPtr = artArray.data.add(i * info.artArrayEntrySize);
           if (fieldPtr.equals(needle)) {
             return true;
--- a/lib/android.js
+++ b/lib/android.js
@@ -967,14 +967,14 @@ export function getArtClassSpec (vm) {
     const fInfo = {
       artArrayLengthSize: 4,
       artArrayEntrySize: fieldSpec.size,
-      // java/lang/Integer has 11 fields on Android 16.
-      artArrayMax: 20
+      // java/io/File has 15 fields on Android 16.
+      artArrayMax: 25
     };
 
     const mInfo = {
       artArrayLengthSize: pointerSize,
       artArrayEntrySize: methodSpec.size,
-      // java/lang/Integer has 66 methods on Android 16.
+      // java/io/File has 63 methods on Android 16.
       artArrayMax: 100
     };
 
@@ -1014,7 +1014,7 @@ export function getArtClassSpec (vm) {
       return false;
     };
 
-    const clazz = env.findClass('java/lang/Integer');
+    const clazz = env.findClass('java/io/File');
     const clazzRef = env.newGlobalRef(clazz);
 
     try {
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks!

> Edit: Also added an extra check for the art array length. Sometimes it would read a negative number when trying offsets, and it would infinitely loop.

This seems impossible looking at the code, given that `readU*` return unsigned values. Would be good to get to the bottom of why, so we don't paper over a bug that might bite us later.

oleavr:
Thanks! :raised_hands:

ExternalAddress4401:
Sorry to be this guy but I'm trying to track down one of 2 things that might be causing my issues.

With this change my error is now 

`07-21 20:02:37.145 27126 27126 F DEBUG   : Abort message: 'No pending exception expected: java.lang.NoSuchFieldError: no "Ljava/lang/String;" field "path" in class "Ljava/io/File;" or its superclasses`

```
import "frida-il2cpp-bridge";
import Logger from "../lib/Logger.js";
import fs from "frida-fs";
import { decrypt } from "../lib/Encrypter.js";
import { Buffer } from "buffer";
import Java from "frida-java-bridge";

Java.perform(() => {
  Logger.log("Inside java");
  Il2Cpp.perform(async () => {
  Logger.log("Inside");
  const path = "sdcard/script/override.js";

  const script: any = fs.readFileSync(path);
  const code = Buffer.from(script.toString(), "base64").toString();
  const decrypted = decrypt(code);

  Logger.log("decrypted");

  try {
    eval(decrypted);
    Logger.log("eval");
  } catch (e) {
    const error = e as Error;
    Logger.log(`Error running script: ${error.message}`);
  }
  }, "main");
});
```

This eval's a simple script

```
import Java from "frida-java-bridge";
import Logger from '../lib/Logger';

Java.perform(() => {
	Logger.log("Hello!");
})
```

Where Logger contains my aforementioned `var Log = Java.use("android.util.Log");` calls. This eval'd script crashes with the error I gave. 

You'll notice this uses [frida-il2cpp-bridge](https://github.com/vfsfitvnm/frida-il2cpp-bridge) and when I comment out the calls to Il2Cpp.Perform it NO LONGER crashes.

I find the error peculiar as it relates to this PR but the commented out code points to the library. I'm going to post this in both places to try and narrow down the issue but I'd appreciate any insight if you have it.

AeonLucid:
@ExternalAddress4401 Can you provide a minimal script, frida cli arguments and a target app so that I can reproduce the error?

ExternalAddress4401:
My setup is somewhat of a mess so I have the smallest repro I can manage at: https://github.com/ExternalAddress4401/frida-repro

1. Clone this repro
2. `npm install`
3. `node build.js` <-- this will create a file.js made from device.ts
4. Copy the contents of file.js into https://www.base64encode.org/ and copy the output
5. Open agent.ts and replace the `const c = ...` with your new base64 string
6. `node build.js -agent` <-- this will build file.js with the contents of agent.ts instead

I use https://beatstar.en.uptodown.com/android/download as a test. I think this should work for any game that uses Il2CPP though.

7. Start the game with `frida -U -f ... -l file.js` (If you're using the beatstar APK then `frida -U -f com.spaceapegames.beatstar -l file.js`)
8. Observe the CLI log

```
[SM G781W::Gadget ]-> Error: java.lang.NoSuchFieldError: no "Ljava/lang/String;" field "path" in class "Ljava/io/File;" or its superclasses
    at <anonymous> (C:\Users\megam\Desktop\repo\file.js:5603)
    at value (C:\Users\megam\Desktop\repo\file.js:16081)
    at m (C:\Users\megam\Desktop\repo\file.js:14522)
    at apply (native)
    at value (C:\Users\megam\Desktop\repo\file.js:15932)
    at m (C:\Users\megam\Desktop\repo\file.js:14469)
    at <anonymous> (C:\Users\megam\Desktop\repo\file.js:16813)
    at <anonymous> (C:\Users\megam\Desktop\repo\file.js:3017)
    at call (native)
    at <anonymous> (C:\Users\megam\Desktop\repo\file.js:2552)
```

* You don't actually **_have_** to use `-f` you can start an application and just supply the name of the process as well and it doesn't make a different apparently.

AeonLucid:
@ExternalAddress4401 Thanks! I forgot to ask but which device and android version is this on? Any special roms? 

If possible you can also message me on discord or telegram (both are @AeonLucid).

ExternalAddress4401:
> @ExternalAddress4401 Thanks! I forgot to ask but which device and android version is this on? Any special roms?
> 
> If possible you can also message me on discord or telegram (both are @AeonLucid).

Nope this is a stock ROM and a Samsung Galaxy S20 Canadian variant so it's a locker bootloader and cannot be rooted or anything of that kind.

AeonLucid:
> > @ExternalAddress4401 Thanks! I forgot to ask but which device and android version is this on? Any special roms?
> > If possible you can also message me on discord or telegram (both are @AeonLucid).
> 
> Nope this is a stock ROM and a Samsung Galaxy S20 Canadian variant so it's a locker bootloader and cannot be rooted or anything of that kind.

Thanks for the repro repository. I managed to fix it.

```
[A015::com.spaceapegames.beatstar ]-> [repro] doing
[repro] inside
[repro] doing
[repro] inside
[A015::com.spaceapegames.beatstar ]-> [repro] got code
[repro] (() => {
 
[repro] hello from eval
[repro] eval
```

Edit: https://github.com/frida/frida-java-bridge/pull/362

